PR #3, opened 2023-10-19
OEL-2673: Open a subscription modal when clicking on the link

AaronGilMartinez:
Modal form


modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.libraries.yml
```diff
@@ -0,0 +1,4 @@
+oe_subscriptions_anonymous:
```
brummbar: Why add a library with one dependency only? Might as well add that library to the modal.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.libraries.yml
```diff
@@ -0,0 +1,4 @@
...
+  version: 1.x
+  dependencies:
+    - core/drupal.dialog.ajax
```
brummbar: The correct dependency is core/drupal.ajax.

modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
```diff
@@ -47,9 +49,8 @@ public static function create(ContainerInterface $container) {
    * {@inheritdoc}
    */
   public function subscribeAnonymous(string $subscription_id) {
-    $response = new AjaxResponse();
-    $response->addCommand(new AlertCommand("Subscription ID: $subscription_id"));
-    return $response;
+    // Validation?
```
brummbar: Any comment that was in the poc must be adressed. In this case, validation means we need to check if the subscription id is valid before generating the form.
We need to:
1. Remove the wrapper controller and add the form in the route directly.
2. Add a custom access callback to validate that the link is valid.

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
```diff
@@ -88,6 +88,11 @@ public function view(ContentEntityInterface $entity) {
         'class' => ['use-ajax', 'button', 'button--small'],
         'data-dialog-type' => 'modal',
       ],
+      '#attached' => [
+        'library' => [
+          'oe_subscriptions_anonymous/oe_subscriptions_anonymous',
```
brummbar: This is where you can add `core/drupal.ajax`.

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
```diff
@@ -88,6 +88,11 @@ public function view(ContentEntityInterface $entity) {
...
+        ],
+      ],
```
brummbar, line 98: Now we need to generate the correct link. The current one is hardcoded to a specific flag.
The subscrition id can be removed, maybe we just pass 2 params, flag id and entity id. We can determine the entity type id from the flag itself.
We need to handle the caching in line 79 too.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,127 @@
...
+   * @var \Drupal\flag\FlagInterface
+   */
+  protected $flag;
+
+  /**
+   * The node.
+   *
+   * @var \Drupal\node\NodeInterface
+   */
+  protected $node;
```
brummbar: This is not needed. Do not create entities in the setup unless they are shared between all methods. There's only one method now.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,127 @@
...
+      'name' => 'Article',
+    ]);
+    // Create a flag.
+    $this->flag = $this->createFlagFromArray([
+      'id' => 'subscribe_article',
+      'label' => 'Subscribe article',
+      'entity_type' => 'node',
+      'bundles' => ['article'],
+      'link_type' => 'reload',
+      'global' => FALSE,
```
brummbar: See kernel test, need to pass different values, some can be omitted.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,127 @@
...
+    ]);
+    // Create the node.
+    $this->node = Node::create([
```
brummbar: Use `$this->drupalCreateNode()`.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,127 @@
...
+    $page = $session->getPage();
+    $assert_session = $this->assertSession();
+    $form_selector = "form.oe-subscriptions-anonymous-subscribe-form";
+    // Got to node page with subscription.
+    $this->drupalGet('node/' . $this->node->id());
+    // Click subscribe link.
+    $this->clickLink('Anonymous Subscribe');
+    $assert_session->assertWaitOnAjaxRequest();
+    // The form is shown.
+    $assert_session->elementExists('css', $form_selector);
```
brummbar: Assert modal rather than form.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,127 @@
...
+    // E-mail.
+    $email = $assert_session->fieldEnabled('email');
+    $this->assertTrue($email->isVisible());
+    $this->assertTrue($email->hasAttribute('required'));
+    // Terms and conditions.
+    $terms = $assert_session->fieldEnabled('accept_terms');
```
brummbar: Use labels not machine names.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,127 @@
...
+    $this->assertTrue($terms->isVisible());
```
brummbar: Why this check? It's enough to assert that the field is there. It has to be visible to click it later.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,127 @@
...
+    $this->assertTrue($terms->hasAttribute('required'));
+    // Suscribe.
+    $subscribe = $page->find('css', 'button.button.form-submit');
```
brummbar: Cache the modal in a variable and find the elements inside it.
Test that fields are required by pressing the button without filling them in.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,127 @@
...
+    $this->assertTrue($subscribe->isVisible());
+    // No thanks.
+    $cancel = $page->find('css', 'button.button.dialog-cancel');
+    $this->assertTrue($cancel->isVisible());
+    // Close modal.
+    $close = $page->find('css', 'button.ui-dialog-titlebar-close');
+    $close->click();
```
brummbar: Use the cancel button to tst that it closes, and check that no messages have been added to the page. This can be done after filling the data, so we test that the form doesn't submit for real.

brummbar (changes requested):

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.libraries.yml
AaronGilMartinez: Library removed, drupal.ajax added to the modal too.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.libraries.yml
AaronGilMartinez: Fixed

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
AaronGilMartinez: - Changed routing to form instead of controller.
- Added access checker service to validate data

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
AaronGilMartinez: - Changed to 2 parameters flag_id:entity_id from deriver
- Added cache to link

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Moved to the test

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Removed unnecessary parameters

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
```diff
@@ -59,36 +60,44 @@ public static function create(ContainerInterface $container, array $configuratio
...
-    // Based on entity available flags.
-    $flags = $this->flag->getAllFlags($entity_type, $entity->bundle);
-    // No flaggings to subscribe through.
-    if (empty($flags)) {
+    $cache = new CacheableMetadata();
+    // Based on derivative id.
+    $flag = $this->flag->getFlagById($this->getDerivativeId());
+    // No flag empty return.
+    if (empty($flag)) {
+      $cache->applyTo($build);
```
brummbar: We are applying an empty cache here, so it's not needed.

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
```diff
@@ -59,36 +60,44 @@ public static function create(ContainerInterface $container, array $configuratio
...
       return $build;
     }
-    // Get 'Susbscribe to' controller method.
+    // Get link to form.
     $url = Url::fromRoute('oe_subscriptions_anonymous.anonymous_subscribe', [
       'subscription_id' => implode(':', [
```
brummbar: With two parameters I meant a url like `/subscribe/{flag}/{entity_id}`. Like it's in flag route.

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
```diff
@@ -59,36 +60,44 @@ public static function create(ContainerInterface $container, array $configuratio
...
-        'subscribe_node',
-        $entity_type,
+        $flag->id(),
         $entity->id(),
       ]),
     ]);
+    // Cache based on flag.
+    $cache->addCacheableDependency($flag);
     // No access.
     if (!$url->access()) {
```
brummbar: We should get the result as object and add it to the cacheable dependencies.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.routing.yml
```diff
@@ -2,7 +2,7 @@ oe_subscriptions_anonymous.anonymous_subscribe:
   path: '/subscribe/{subscription_id}'
   defaults:
     _title: 'Subscribe'
-    _controller: '\Drupal\oe_subscriptions_anonymous\Controller\SubscriptionAnonymousController::subscribeAnonymous'
+    _form: '\Drupal\oe_subscriptions_anonymous\Form\AnonymousSubscribeForm'
   requirements:
     _user_is_logged_in: 'FALSE'
-    _access: 'TRUE'
+    _subscription_access_check: 'TRUE'
```
brummbar: Another way was to use `_custom_access` and pass a callback, but we'll keep it this way.
Let's make it a bit more namespaced: `_anonymous_subscription_access`.

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
```diff
@@ -59,36 +60,44 @@ public static function create(ContainerInterface $container, array $configuratio
...
-      // @todo Handle caching.
+      $cache->applyTo($build);
       return $build;
     }
     // Link.
     $build = [
       '#type' => 'link',
-      '#title' => $this->t('Anonymous Subscribe'),
+      '#title' => $flag->label(),
```
brummbar: This now shows like this 
image
We should use the subscribe label configured in the flag rather than the flag name.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.routing.yml
```diff
@@ -2,7 +2,7 @@ oe_subscriptions_anonymous.anonymous_subscribe:
   path: '/subscribe/{subscription_id}'
```
brummbar: By splitting this parameter, and using {flag} we get an upcasted entity.

modules/oe_subscriptions_anonymous/src/Access/SubscriptionAccessCheck.php
```diff
@@ -0,0 +1,86 @@
...
+   * Anonymous subscription access check.
+   *
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
+   *   The route matched.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(RouteMatchInterface $route_match): AccessResultInterface {
+    $subscription_id = $route_match->getParameter('subscription_id');
```
brummbar: In here we will be able to load the entity to be flagged by using `$flaggable = $this->flagService->getFlaggableById($flag, $flaggable_id);`
We should then check if the anon user can view the node.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -0,0 +1,122 @@
...
+    $form['#attached']['library'][] = 'core/drupal.ajax';
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $this->messenger()->addMessage($this->t('A confirmation e-email has been sent to your e-mail address.'));
```
brummbar, line 66: We will make a followup to return js commands when on a modal, because the redirect is not handled.

modules/oe_subscriptions_anonymous/tests/src/Functional/AnonymousLinkTest.php
```diff
@@ -99,36 +96,47 @@ protected function setUp(): void {
...
+          'format' => filter_default_format(),
+        ],
+      ],
+      'type' => 'page',
+      'title' => $this->randomMachineName(8),
+      'uid' => $adminUser->id(),
+      'status' => 1,
+      'promote' => 0,
+      'sticky' => 0,
+    ]);
```
brummbar: Why add all these field values here?

1. Most of them are already added by the drupalCreateNode() call.
2. Most of them are not needed for the test (promote, sticky) and a reader has a harder time to understand that, since it seems that value will impact the test.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+        ],
+      ],
+      'type' => 'article',
+      'title' => $this->randomMachineName(8),
+      'uid' => 0,
+      'status' => 1,
+      'promote' => 0,
+      'sticky' => 0,
+    ]);
+    $node->save();
```
brummbar:
```suggestion
    $node = $this->drupalCreateNode([
      'type' => 'article',
      'status' => 1,
    ]);
```

modules/oe_subscriptions_anonymous/tests/src/Functional/AnonymousLinkTest.php
```diff
@@ -4,7 +4,6 @@
```
brummbar: Let's delete this test, there's quite a mess here. E.g.
```
$adminUser = $this->createUser([
      'administer flags',
      'administer flagging display',
      'administer flagging fields',
      'administer node display',
      'administer modules',
      'administer nodes',
      'create article content',
      'edit any article content',
      'delete any article content',
    ]);
```
I doubt that the permission to administer modules is needed.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+ */
+class SubscribeModalFormTest extends WebDriverTestBase {
+
+  use FlagCreateTrait;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'system',
```
brummbar: Remove unneded dependencies.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+      'type' => 'article',
+      'title' => $this->randomMachineName(8),
+      'uid' => 0,
+      'status' => 1,
+      'promote' => 0,
+      'sticky' => 0,
+    ]);
+    $node->save();
+
+    $session = $this->getSession();
```
brummbar: Why do you need $session?

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+    $page = $session->getPage();
+    $assert_session = $this->assertSession();
+    // Using ui-dialog parent of modal-form,
+    // Submit buttons and close in siblings.
+    $modal_select = '.ui-dialog';
+    // Got to node page with subscription.
+    $this->drupalGet('node/' . $node->id());
```
brummbar:
```suggestion
    $this->drupalGet($page->toUrl());
```

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+    // Click subscribe link.
+    $this->clickLink($flag->label());
+    $assert_session->assertWaitOnAjaxRequest();
+    // The modal wrapper.
+    $modal = $page->find('css', $modal_select);
+    // Find all elements.
+    $mail_label = 'Your e-mail';
+    $terms_label = 'I have read and agree with the data protection terms.';
+    // Using xpath, findButton get input (hidden) instead button tags.
+    $mail_field = $modal->findField($mail_label);
```
brummbar:
```suggestion
    $mail_field = $assert_session->fieldExists($mail_label, $modal);
```

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -0,0 +1,122 @@
...
+      'cancel' => [
+        '#type' => 'button',
+        '#value' => $this->t('No thanks') ,
+        '#attributes' => [
+          'class' => ['dialog-cancel'],
+        ],
+      ],
+      'submit' => [
+        '#type' => 'submit',
+        '#value' => $this->t('Subscribe me'),
```
brummbar, line 53: Let's add `'#button_type' => 'primary',` so it shows in a different color.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+      'sticky' => 0,
+    ]);
+    $node->save();
+
+    $session = $this->getSession();
+    $page = $session->getPage();
+    $assert_session = $this->assertSession();
+    // Using ui-dialog parent of modal-form,
+    // Submit buttons and close in siblings.
+    $modal_select = '.ui-dialog';
```
brummbar: This is not a select, it's a selector.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+    $modal = $page->find('css', $modal_select);
+    // Find all elements.
+    $mail_label = 'Your e-mail';
+    $terms_label = 'I have read and agree with the data protection terms.';
+    // Using xpath, findButton get input (hidden) instead button tags.
+    $mail_field = $modal->findField($mail_label);
+    $terms_field = $modal->findField($terms_label);
+    // Using xpath, findButton get input (hidden) instead button tags.
+    // Submit has no specific class to select thourgh CSS.
+    $subscribe = $modal->find('xpath', '//button[contains(text(),"Subscribe")]');
```
brummbar: The button doesn't have "Subscribe" as text, it's "Subscribe me".

brummbar: There's no need to collect elements in variables, especially if used once.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+    $cancel = $modal->find('xpath', '//button[contains(text(),"No thanks")]');
+    $close = $modal->find('css', 'button.ui-dialog-titlebar-close');
+    // Try to submit, empty form.
+    $this->disableNativeBrowserRequiredFieldValidation();
+    $subscribe->press();
```
brummbar: Xpath can be replaced by this:
```
$button_pane = $assert_session->elementExists('css', '.ui-dialog-buttonpane', $modal);
$assert_session->buttonExists('Subscribe me', $button_pane)->press();
```

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+    $this->assertSession()->pageTextContains("$mail_label field is required.");
+    $this->assertSession()->pageTextContains("$terms_label field is required.");
+    // Test Close modal button.
+    $this->drupalGet('node/' . $node->id());
+    $this->clickLink($flag->label());
+    $assert_session->assertWaitOnAjaxRequest();
+    $close->press();
+    $assert_session->elementNotExists('css', $modal_select);
```
brummbar: This should be also here `$assert_session->statusMessageNotExists();`.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,119 @@
...
+    // Test cancel button, filled data.
+    $this->clickLink($flag->label());
+    $assert_session->assertWaitOnAjaxRequest();
+    $mail_field->setValue('test@test.com');
+    $terms_field->check();
+    $cancel->press();
+    $assert_session->elementNotExists('css', $modal_select);
+    // No message no submit.
+    $assert_session->statusMessageNotExists();
```
brummbar, line 95: No tests with correct submission?

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
```diff
@@ -0,0 +1,192 @@
...
+use Drupal\Core\Routing\RouteMatch;
+use Drupal\entity_test\Entity\EntityTestBundle;
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\node\Entity\Node;
+use Drupal\Tests\flag\Traits\FlagCreateTrait;
+
+/**
+ * Tests the access checker.
+ */
+class AccessCheckTest extends KernelTestBase {
```
brummbar, line 19: This type of test, using directly the access check itself, is more suitable for testing an access check class rather than access in a specific route.
If I replace the $route with `$route = new \Symfony\Component\Routing\Route('/anything/{subscription_id}');`, the test is still green.
It doesn't really test that our route has the access check set, it only tests that the route contains `{subscription_id}`.
Let's change the test to use \Drupal\Core\Access\AccessManager::checkNamedRoute(). This allows to run the checks directly on the real route.

brummbar (changes requested):

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
AaronGilMartinez: Not sure to understand this, you mean to create a link object?

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.routing.yml
AaronGilMartinez: Renamed

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
AaronGilMartinez: Replaced with flag short text

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Corrected, removed one use variables

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Replaced

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.routing.yml
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/src/Access/SubscriptionAccessCheck.php
AaronGilMartinez: Done, removed entity manager too, not needed then

modules/oe_subscriptions_anonymous/tests/src/Functional/AnonymousLinkTest.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Thanks

modules/oe_subscriptions_anonymous/tests/src/Functional/AnonymousLinkTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Renamed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Added one for the moment, until next task it's done shows the message when sutmited. I can add more if needed.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
AaronGilMartinez: Ok

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
AaronGilMartinez: Test changed to use access_manager. I had to apply a patch to get the test working: https://www.drupal.org/project/drupal/issues/3158130

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
AaronGilMartinez: Clarified, its the url::access method, get object to add cache dependency

composer.json
```diff
@@ -53,6 +54,11 @@
             "build/modules/contrib/{$name}": ["type:drupal-module"],
             "build/themes/contrib/{$name}": ["type:drupal-theme"]
         },
+        "patches": {
+            "drupal/core": {
+                "https://www.drupal.org/project/drupal/issues/3158130": "https://www.drupal.org/files/issues/2022-12-11/3158130-26.patch"
```
brummbar: Format for patches is
`"<issue title> @see <issue link>": "<patch url>"`.

modules/oe_subscriptions_anonymous/src/Access/SubscriptionAccessCheck.php
```diff
@@ -0,0 +1,65 @@
...
+   */
+  protected FlagServiceInterface $flagService;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(
+    FlagServiceInterface $flagService) {
+    $this->flagService = $flagService;
+  }
```
brummbar: No need to break indent. If we are breaking indentation, then the format is
```
public function __construct(
  FlagServiceInterface $flagService
) {
  $this->flagService = $flagService;
}

File: modules/oe_subscriptions_anonymous/src/Access/SubscriptionAccessCheck.php [hunk 38]
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
+   *   The route matched.
+   * @param \Drupal\flag\FlagInterface $flag
+   *   The flag to subscribe to.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(RouteMatchInterface $route_match, FlagInterface $flag = NULL): AccessResultInterface {
```
brummbar: Why $flag = NULL?
This makes the parameter optional, but our parameter should always be there.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -0,0 +1,126 @@
...
+   *   The renderer.
+   */
+  public function __construct(
+    MailManagerInterface $mailManager,
+    ConfigFactoryInterface $configFactory,
+    RendererInterface $renderer) {
+    $this->mailManager = $mailManager;
+    $this->configFactory = $configFactory;
+    $this->renderer = $renderer;
+  }
```
brummbar: Same as above. I also missed that we don't need any of these services at all.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, FlagInterface $flag = NULL, $entity_id = NULL) {
+    $form_state->set('flag', $flag);
+    $form_state->set('entity_id', $entity_id);
+    // Default Text without link.
+    $title = $this->t('I have read and agree with the data protection terms.');
```
brummbar: Why a variable?

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
```diff
@@ -59,36 +60,44 @@ public static function create(ContainerInterface $container, array $configuratio
...
     $build = [];
-    $entity_type = $entity->getEntityTypeId();
-    // Based on entity available flags.
-    $flags = $this->flag->getAllFlags($entity_type, $entity->bundle);
-    // No flaggings to subscribe through.
-    if (empty($flags)) {
+    $cache = new CacheableMetadata();
+    // Based on derivative id.
+    $flag = $this->flag->getFlagById($this->getDerivativeId());
+    // No flag empty return.
```
brummbar: A good comment would be "Bail out if the flag cannot be loaded."

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $node = $this->drupalCreateNode([
+      'type' => 'article',
+      'status' => 1,
+    ]);
+    $node->save();
+
+    $page = $this->getSession()->getPage();
+    $assert_session = $this->assertSession();
+    // Using ui-dialog parent of modal-form,
+    // Submit buttons and close in siblings.
```
brummbar, line 57: I am not sure I understand this.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $modal_selector = '.ui-dialog';
+
+    // Got to node page with subscription.
```
brummbar: No need for this comment.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $this->drupalGet($node->toUrl());
+    // Click subscribe link.
+    $link_text = $flag->getShortText('flag');
```
brummbar: You set a defined value on top, you shouldn't use the flag method here. You are expecting that text on top to be used for the flag. So you can safely use "Subscribe article" (even better if we change it to a realistic "Subscribe to this article").

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $this->clickLink($link_text);
+    $assert_session->assertWaitOnAjaxRequest();
+
+    // The modal wrapper.
+    $modal = $page->find('css', $modal_selector);
```
brummbar: In general, use assert_session->elementExists().

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $modal_selector = '.ui-dialog';
+
+    // Got to node page with subscription.
+    $this->drupalGet($node->toUrl());
+    // Click subscribe link.
+    $link_text = $flag->getShortText('flag');
+    $this->clickLink($link_text);
+    $assert_session->assertWaitOnAjaxRequest();
+
+    // The modal wrapper.
```
brummbar, line 64: This is not needed.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $modal = $page->find('css', $modal_selector);
+    // Find all elements.
```
brummbar: We are asserting the fields of the modal. So better a "Assert the fields of the modal."

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $mail_label = 'Your e-mail';
+    $terms_label = 'I have read and agree with the data protection terms.';
+    $mail_field = $assert_session->fieldExists($mail_label, $modal);
+    $terms_field = $assert_session->fieldExists($terms_label, $modal);
+    // Get button pane with submit buttons.
+    $button_pane = $assert_session->elementExists('css', '.ui-dialog-buttonpane', $modal);
+
+    // Try to submit, empty form.
```
brummbar:
```suggestion
    // Verify that mail and terms field are marked as required.
```

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $this->disableNativeBrowserRequiredFieldValidation();
+    $assert_session->buttonExists('Subscribe me', $button_pane)->press();
+    $this->assertSession()->pageTextContains("$mail_label field is required.");
+    $this->assertSession()->pageTextContains("$terms_label field is required.");
+
+    // Test Close modal button.
```
brummbar: No need for caps.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -0,0 +1,118 @@
...
+    $assert_session->statusMessageNotExists();
+
+    // Test submit.
+    $this->clickLink($link_text);
+    $assert_session->assertWaitOnAjaxRequest();
+    $mail_field->setValue('test@test.com');
+    $terms_field->check();
+    $assert_session->buttonExists('Subscribe me', $button_pane)->press();
+    $assert_session->elementNotExists('css', $modal_selector);
+    $assert_session->statusMessageExists();
```
brummbar: Now we need to test that the confirmation message is shown.

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
```diff
@@ -0,0 +1,185 @@
...
+    $node = $this->createNode([
+      'type' => 'article',
+      'status' => 1,
+    ]);
+    $node->save();
+    $nid = $node->id();
+    $access_manager = $this->container->get('access_manager');
+    // Give access content permission to anonymous.
+    $this->grantPermissions(Role::load('anonymous'), ['access content']);
+    $this->setCurrentUser(new AnonymousUserSession());
```
brummbar: We finished setup now, add an empty line below this.

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
```diff
@@ -0,0 +1,185 @@
...
+    // Empty.
```
brummbar: Comment not good, maybe // Empty parameter values. ?

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
```diff
@@ -0,0 +1,185 @@
...
+      [
+        'flag' => $flag_id,
+        'entity_id' => '1234',
+      ],
+      NULL,
+      TRUE,
+    );
+    $this->assertFalse($access->isAllowed());
+
+    // Finally a subscribe parameters matching all conditions.
```
brummbar: Incorrect english.

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
```diff
@@ -0,0 +1,185 @@
...
+      [
+        'flag' => $flag_id,
+        'entity_id' => $nid,
+      ],
+      $this->createUser([], 'logged_user'),
+      TRUE,
+    );
+    $this->assertFalse($access->isAllowed());
+
+  }
```
brummbar, line 166: Let's add an helper method to this test to improve readability.
`protected function assertRouteAccess(string $flag_id, string $entity_id, AccountInterface $user = NULL)`. So you just pass the minimum needed information and it becomes more readable.

brummbar (changes requested):
Last minor things to be fixed.

composer.json
AaronGilMartinez: Fixed

modules/oe_subscriptions_anonymous/src/Access/SubscriptionAccessCheck.php
AaronGilMartinez: Fixed

modules/oe_subscriptions_anonymous/src/Access/SubscriptionAccessCheck.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/src/Plugin/ExtraField/Display/AnonymousSubscribeLink.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: The form has ui-dialog as parent, and then modal-fom and ui-dialog-buttonpane as children at the same level, just wanted to clarify I couldn't use modal-form.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Modal is not present if we try to check that exists just after, I have changed the function to
 waitForElement

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
AaronGilMartinez: Fixed

modules/oe_subscriptions_anonymous/tests/src/Kernel/AccessCheckTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
AaronGilMartinez: Removed

brummbar (approved):

PR #4, opened 2023-11-22
OEL-2674: Implement email validation for the subscription link

AaronGilMartinez:


modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.install
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Implements function hook_schema().
+ */
+function oe_subscriptions_anonymous_schema() {
+  $schema['oe_subscriptions_anonymous_subscriptions'] = [
+    'description' => 'Stores anonymous subscription in order to manage validation',
+    'fields' => [
+      'mail' => [
+        'description' => 'Subscription mail.',
+        'length' => 255,
```
brummbar: Let's use the same value as \Drupal\Core\Field\Plugin\Field\FieldType\EmailItem::schema .

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.install
```diff
@@ -0,0 +1,58 @@
...
+        'type' => 'varchar_ascii',
+        'not null' => TRUE,
+        'default' => '',
```
brummbar: Is there a reason for empty string as default?

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.install
```diff
@@ -0,0 +1,58 @@
...
+      ],
+      'hash' => [
+        'description' => 'Temporary validation hash.',
+        'length' => 255,
+        'type' => 'varchar_ascii',
+        'not null' => TRUE,
+      ],
+      'changed' => [
+        'description' => 'Last time the subscription was changed.',
+        'length' => 255,
```
brummbar: Int is the correct type.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.routing.yml
```diff
@@ -6,3 +6,19 @@ oe_subscriptions_anonymous.anonymous_subscribe:
   requirements:
     _user_is_logged_in: 'FALSE'
     _anonymous_subscription_access: 'TRUE'
+
+oe_subscriptions_anonymous.anonymous_confirm:
+  path: '/confirm/{flag}/{entity_id}/{email}/{hash}'
```
brummbar: These routes are too generic, we might clash with other modules. We need to pick a starting point for these.
Let's have them start with /subscribe/, so /subscribe/confirm, subscribe/cancel, subscribe/request

modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
```diff
@@ -0,0 +1,248 @@
...
+    $this->flagService = $flagService;
+    $this->mailValidatorService = $mailValidatorService;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function createSubscription(string $mail, FlagInterface $flag, string $entity_id): string {
+    // Validate mail.
+    if (!$this->mailValidatorService->isValid($mail)) {
```
brummbar: It's not the job of this service to validate the email.

modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
```diff
@@ -0,0 +1,248 @@
...
+use Drupal\Component\Utility\EmailValidatorInterface;
+use Drupal\Core\Database\Connection;
+use Drupal\flag\FlagInterface;
+use Drupal\flag\FlagServiceInterface;
+use Drupal\user\Entity\User;
+
+/**
+ * Class to manage anonymous subscriptions.
+ */
+class AnonymousSubscriptionManager implements AnonymousSubscriptionManagerInterface {
```
brummbar, line 15: This needs to be split in two services.
One service will handle tokens. We will need this later also for allowing anon users to access their subscriptions page. It should have methods like:

1. `isValid(string $token, string $email, string $scope)`: checks if a token exists and is valid (not expired).
2. `get(string $email, string $scope)`: returns a token. It will create the entry if needed, or updated it.
3. `delete(string $email, string $scope)`: deletes a token given some parameters.
4. `deleteExpired()`: deletes all expired tokens.

For `$scope` above, we need to choose what to do:
- we have a single service for the subscription process and the subscriptions edit page. In this case, we build the scope using a static method in an helper class.
- we have a service per token type, but we need to have double interfaces and all.

modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
```diff
@@ -0,0 +1,248 @@
...
+      return FALSE;
+    }
+
+    // Try to load user.
+    $account = user_load_by_mail($mail);
+
+    // Create decoupled user.
+    if ($account === FALSE) {
+      $user = User::create(['mail' => $mail])->save();
+      $account = User::load($user);
```
brummbar: Always use dependency injection.

modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
brummbar: The other service will be the AnonymousSubscriptionManager, which should just take care of subscribing anonymous users. The method for now should be just `subscribe(string $email, FlagInterface $flag, string $entity_id)`.
The validation of the hash is to be done in the controller (like core does).

brummbar (changes requested):
Let's start with these changes.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.install
AaronGilMartinez: Added

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.install
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.install
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.routing.yml
AaronGilMartinez: Added

modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
AaronGilMartinez: Service splitted, using a single service for the scope.

PR #11, opened 2023-12-12
OEL-2761: Style the "confirm subscription" message

AaronGilMartinez:


modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -164,7 +169,14 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       return;
     }
 
-    $this->messenger()->addMessage($this->t('A confirmation e-email has been sent to your e-mail address.'));
+    $confirm_message = [
+      '#theme' => 'oe_subscriptions_anonymous_message',
+      '#title' => $this->t('A confirmation email has been sent to your email address'),
+      '#message' => $this->t('To confirm your subscription, <strong>please click on the confirmation link</strong> sent to your e-mail address.'),
```
brummbar: Instead of having the strings here, let's have everything in the template.

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
```diff
@@ -163,7 +165,8 @@ public function testForm(): void {
     $assert_session->fieldExists($mail_label)->setValue('another@example.com');
     $assert_session->fieldExists($terms_label)->check();
     $assert_session->buttonExists('Subscribe me')->press();
-    $assert_session->statusMessageContains('A confirmation e-email has been sent to your e-mail address.', 'status');
+    $assert_session->statusMessageContains('A confirmation email has been sent to your email address', 'warning');
+    $assert_session->statusMessageContains('To confirm your subscription, please click on the confirmation link sent to your e-mail address.', 'warning');
```
brummbar: We need to somehow test that the message contains the html. We cannot use this.

brummbar (changes requested):

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
AaronGilMartinez: Moved

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
AaronGilMartinez: Added a function to check tag => content inside the warning message

modules/oe_subscriptions_anonymous/tests/src/Trait/MessageTemplateTrait.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_subscriptions_anonymous\Trait;
+
+/**
+ * Trait to assert messages templates.
+ */
+trait MessageTemplateTrait {
```
brummbar: Why message template? Message templates are for the email sent, not the confirmation messages.

modules/oe_subscriptions_anonymous/tests/src/Trait/MessageTemplateTrait.php
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Asserts that the confirm message is present.
+   *
+   * @return bool
+   *   If the message is present in the page.
+   */
+  public function confirMessageExists(): bool {
```
brummbar: 1. Methods that assert stuff (see first line of the phpdoc) should start with `assert`.
2. Method should be protected.
3. Parameters should not be hardcoded, we should pass an array of values.
4. Method should be called assertHtmlStatusMessage()
5. The message type should be passed too.

modules/oe_subscriptions_anonymous/tests/src/Trait/MessageTemplateTrait.php
```diff
@@ -0,0 +1,33 @@
...
+      'p' => 'To confirm your subscription, please click on the confirmation link sent to your e-mail address.',
+      'strong' => 'please click on the confirmation link',
+    ];
+    // All elements in the array have to be present.
+    $matches = array_filter($message, function ($value, $key) {
+      $selector = '//div[@data-drupal-messages]//div[contains(@aria-label, "Warning message") or contains(@aria-labelledby, "warning")]//' . $key . '[contains(.,"' . $value . '")]';
+      return $this->assertSession()->elementExists('xpath', $selector);
+    }, ARRAY_FILTER_USE_BOTH);
+
+    return empty($matches);
```
brummbar: Why overcomplicate the method with an array filter, and a return value? We do not need the return, and anyway if one of the conditions is not matched, the `elementExists()` method will terminate the test.

modules/oe_subscriptions_anonymous/tests/src/Trait/MessageTemplateTrait.php
```diff
@@ -0,0 +1,33 @@
...
+   */
+  public function confirMessageExists(): bool {
+    $message = [
+      'h4' => 'A confirmation email has been sent to your email address',
+      'p' => 'To confirm your subscription, please click on the confirmation link sent to your e-mail address.',
+      'strong' => 'please click on the confirmation link',
+    ];
+    // All elements in the array have to be present.
+    $matches = array_filter($message, function ($value, $key) {
+      $selector = '//div[@data-drupal-messages]//div[contains(@aria-label, "Warning message") or contains(@aria-labelledby, "warning")]//' . $key . '[contains(.,"' . $value . '")]';
```
brummbar: Use the `buildXPathQuery()` of the assert session. Also cache the assert session itself.

brummbar (changes requested):

modules/oe_subscriptions_anonymous/tests/src/Trait/MessageTemplateTrait.php
AaronGilMartinez: Renamed

modules/oe_subscriptions_anonymous/tests/src/Trait/MessageTemplateTrait.php
AaronGilMartinez: Fixed different issues like visibility, name and parameters

modules/oe_subscriptions_anonymous/tests/src/Trait/MessageTemplateTrait.php
AaronGilMartinez: Return removed

modules/oe_subscriptions_anonymous/tests/src/Trait/MessageTemplateTrait.php
AaronGilMartinez: Done, thanks!

brummbar (approved):

PR #13, opened 2023-12-14
OEL-2765: Allow anonymous users to manage their subscriptions.

brummbar:


modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.routing.yml
```diff
@@ -22,3 +24,25 @@ oe_subscriptions_anonymous.subscription_request.cancel:
   requirements:
     _user_is_logged_in: 'FALSE'
     _anonymous_subscription_access: 'TRUE'
+  options:
+    no_cache: TRUE
+
+oe_subscriptions_anonymous.user_subscriptions.request_access:
+  path: '/user/subscriptions'
```
AaronGilMartinez, line 31: Shouldn't this be `'/user/my-subscriptions'`? Described in the ticket.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.routing.yml
```diff
@@ -22,3 +24,25 @@ oe_subscriptions_anonymous.subscription_request.cancel:
...
+  defaults:
+    _form: '\Drupal\oe_subscriptions_anonymous\Form\UserSubscriptionsRequestAccessForm'
+    _title: 'Manage your subscriptions'
+  requirements:
+    _user_is_logged_in: 'FALSE'
+  options:
+    no_cache: TRUE
+
+oe_subscriptions_anonymous.user_subscriptions.view:
+  path: '/user/subscriptions/{email}/{token}'
```
AaronGilMartinez, line 41: Same here subscriptions to my-subscriptions. Described in the ticket.

modules/oe_subscriptions_anonymous/src/Form/UserSubscriptionsRequestAccessForm.php
```diff
@@ -0,0 +1,96 @@
...
+    ];
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $mail = $form_state->getValue('email');
```
AaronGilMartinez, line 77: Any mail can be request access to the page, even if no subscriptions exist. Should we add a check?

oe_subscriptions.routing.yml
```diff
@@ -0,0 +1,12 @@
+oe_subscriptions.user.subscriptions:
+  path: '/user/{user}/subscriptions'
```
AaronGilMartinez, line 2: Same here

AaronGilMartinez (changes requested):

modules/oe_subscriptions_anonymous/src/Form/UserSubscriptionsRequestAccessForm.php
brummbar: Yes, we need to have a different e-mail that is sent to "normal" users which has a link to their profile page.
The route response already does that (so it takes care in case an account is "upgraded" and the user uses a not-yet-expired link).
I felt we were already out of time for the ticket and wanted to open a follow-up.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.routing.yml
brummbar: Yes I changed it to align to the user/UID/subscriptions . I think it's better if the page /user/my-subscriptions, if implemented, would redirect to user/uid/subscriptions for the current user if logged in, or redirect to the form to request a token.

AaronGilMartinez (approved):

PR #14, opened 2023-12-22
OEL-2763: Allow users to select the notification frequency.

AaronGilMartinez:


composer.json
```diff
@@ -75,7 +84,8 @@
             "drupal/core-composer-scaffold": true,
             "phpro/grumphp": true,
             "dealerdirect/phpcodesniffer-composer-installer": true,
-            "phpstan/extension-installer": true
+            "phpstan/extension-installer": true,
+            "php-http/discovery": true
```
brummbar: This needs to be always set to false in our components.

oe_subscriptions.module
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * @file
+ * Subscriptions module.
+ */
+
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Implements hook_ENTITY_insert().
```
brummbar: Replace placeholders like ENTITY with the specific hook (`flag`).

src/Form/UserSubscriptionsForm.php
```diff
@@ -86,6 +94,14 @@ public function buildForm(array $form, FormStateInterface $form_state, UserInter
       '#default_value' => $this->account->getPreferredLangcode(),
     ];
 
+    $form['message_digest'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Notifications frequency'),
+      '#description' => $this->t("The frequency this account will receive notifications is subscribed to."),
```
brummbar: Always use single quotes unless another single quote is used in the string, or interpolation is needed.

brummbar (changes requested):
After further investigation, I think we should create a submodule that integrates with the digest system.
Additionally, the message_digest module also creates a field per flagging, where it stores the digest frequency. This is not yet supported by us, we have a generic frequency, so for now, every time the user changes their notification frequency in the form, we should load all the flagging for that user and change their values too.

composer.json
AaronGilMartinez: Changed

oe_subscriptions.module
AaronGilMartinez: Done

src/Form/UserSubscriptionsForm.php
AaronGilMartinez: Changed

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
```diff
@@ -0,0 +1,86 @@
...
+   * Submit to save user digest preference.
+   *
+   * @param array $form
+   *   The form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   */
+  public function subscriptionsFormSubmit(array $form, FormStateInterface $form_state): void {
+    $message_digest = $form_state->getValue('message_digest');
+    $this->account->set('message_digest', $message_digest === '' ? NULL : $message_digest)->save();
```
AaronGilMartinez: I may have to add a check here to see if the field exists or not.

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,74 @@
+<?php
+
+/**
```
brummbar, line 3: Missing strict type declaration.

composer.json
```diff
@@ -13,7 +13,8 @@
         "drupal/extra_field": "^2.3",
         "drupal/flag": "^4.0@beta",
         "drupal/message": "^1.4",
-        "drupal/message_subscribe": "^1.2"
+        "drupal/message_digest": "^1.3",
```
brummbar: This should be in require-dev as the dependency is needed for an optional submodule.

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,74 @@
...
+ * Subscriptions digest module.
+ */
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\oe_subscriptions_digest\SubscriptionsFormAlter;
+
+/**
+ * Implements hook_flag_insert().
+ */
```
brummbar, line 23: Let's put some documentation on why this code is needed.

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function oe_subscriptions_digest_form_oe_subscriptions_user_subscriptions_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  /** @var \Drupal\user\UserInterface $user */
+  $user = $form_state->getBuildInfo()['args']['0'];
+  \Drupal::classResolver(SubscriptionsFormAlter::class)->subscriptionsFormAlter($form, $user);
```
brummbar: Rather than passing the user, pass the form state.

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
```diff
@@ -0,0 +1,86 @@
...
+   */
+  public function subscriptionsFormAlter(array &$form, $user): void {
+    $this->moduleHandler->loadInclude('message_digest_ui', 'module');
+    $this->account = $user;
+    $message_digest = $user->get('message_digest');
+    $storage_definition = $message_digest->getFieldDefinition()->getFieldStorageDefinition();
+
+    $form['message_digest'] = [
+      '#type' => 'select',
+      '#title' => t('Notifications frequency'),
```
brummbar: Nit: we can use the string translation trait to we do $this->t(). Not that it changes much, but it's more in line with this being a class.

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
brummbar: We can add that check, but then it should be in the alter above.

modules/oe_subscriptions_digest/tests/src/Functional/AnonymousUserDigestTest.php
```diff
@@ -0,0 +1,77 @@
...
+   *
+   * This method will request the URL via the dedicated form.
+   *
+   * @param string $email
+   *   The e-mail of the user.
+   *
+   * @return string
+   *   The URL.
+   */
+  protected function getAnonymousUserSubscriptionsPageUrl(string $email): string {
```
brummbar: Is this duplicated? I didn't open the editor yet but I believe so, in that case we can have a trait.

src/Form/UserSubscriptionsForm.php
```diff
@@ -51,7 +51,7 @@ public static function create(ContainerInterface $container) {
     $instance = new static(
       $container->get('entity_type.manager'),
       $container->get('flag'),
-      $container->get('current_user')
+      $container->get('current_user'),
```
brummbar: This comma shoudln't be here.

brummbar (changes requested):
First part of the review.

composer.json
AaronGilMartinez: Moved

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Added, other missing strict types too.

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Added

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Changed

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
AaronGilMartinez: Added

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
AaronGilMartinez: Added in the hook

modules/oe_subscriptions_digest/tests/src/Functional/AnonymousUserDigestTest.php
AaronGilMartinez: It is, I had some doubts about dupicating code or creating the trait, but didn't want to call code from one submodule in another. I've moved it, makes sense from tests persepective, this isn't going to break the functionality, just test when both are there.

src/Form/UserSubscriptionsForm.php
AaronGilMartinez: Removed

modules/oe_subscriptions_digest/tests/src/Kernel/MailFlagtest.php
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /**
+   * Tests the flag for mails are created from subscribe flag.
+   */
+  public function testFlagMirroring(): void {
+    // Create a flag with wrong prefix.
+    $flagStorage = $this->container->get('entity_type.manager')->getStorage('flag');
+    $this->createFlagFromArray(['id' => 'another_flag']);
+    $this->assertNull($flagStorage->load('email_something'));
```
brummbar: Good test, but let's load all flags to make the assertion. With the current test, if the code is changed and the flag created starts with `mail_`, it wouldn't fail.

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,85 @@
...
+declare(strict_types=1);
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\oe_subscriptions_digest\SubscriptionsFormAlter;
+
+/**
+ * Implements hook_flag_insert().
+ *
+ * Message-subscribe-email uses message-subscribe module to email subscribers
```
brummbar: Nit: use underscores for module names.

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ */
+function oe_subscriptions_digest_form_oe_subscriptions_user_subscriptions_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  /** @var \Drupal\user\UserInterface $user */
+  $user = $form_state->getBuildInfo()['args']['0'];
+  if (!$user->hasField('message_digest')) {
+    return;
+  }
```
brummbar: It's good, but let's keep the logic in the alter.

runner.yml.dist
```diff
@@ -29,6 +29,7 @@ drupal:
     - "./vendor/bin/drush en config_devel -y"
     - "./vendor/bin/drush en oe_subscriptions -y"
     - "./vendor/bin/drush en oe_subscriptions_anonymous -y"
```
AaronGilMartinez, line 31: New version of decoupled auth solves the problem with the install: https://git.drupalcode.org/project/decoupled_auth/-/compare/2.0.0...3.0.1#d18eedf709e2e4ca50781c8fafd2ea1f6e2f5e3d_265_264
The issue is not present anymore

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
```diff
@@ -0,0 +1,89 @@
...
+   * Alters the form to add digest preference.
+   *
+   * @param array $form
+   *   The form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   */
+  public function subscriptionsFormAlter(array &$form, $form_state): void {
+    $this->moduleHandler->loadInclude('message_digest_ui', 'module');
+    $this->account = $form_state->getBuildInfo()['args']['0'];
```
brummbar: The main issue with this is that the class will be stored in the form state as a whole, including a serialised account variable and the services. We can solve this by using \Drupal\Core\DependencyInjection\DependencySerializationTrait , but not for the user.
I usually prefer to avoid referencing the class in a callback, and instead use a static callback.
This means that we need to reload the user from the form state in the submit, but it's not a big deal, we can make a protected static method.

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * Alters the form to add digest preference.
+   *
+   * @param array $form
+   *   The form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   */
+  public function subscriptionsFormAlter(array &$form, $form_state): void {
```
brummbar: Type hint missing.

modules/oe_subscriptions_digest/tests/src/Functional/UserDigestTest.php
```diff
@@ -0,0 +1,191 @@
...
+use Drupal\field\Entity\FieldConfig;
+use Drupal\message_digest\Entity\MessageDigestInterval;
+use Drupal\Tests\BrowserTestBase;
+use Drupal\Tests\flag\Traits\FlagCreateTrait;
+use Drupal\user\UserInterface;
+
+/**
+ * Tests the user subscriptions page.
+ */
+class UserDigestTest extends BrowserTestBase {
```
brummbar: Make a base class like the others. 
For me:
* we make a base class that contains the test methods, and it requires implementing classes to provide a few methods;
* we make a base class without any test method but only the `doX` methods, and we leave the implementations of the class to add test methods. This is what we did for the other tests.

It's up to you, normally the best scenario would be point 1.

modules/oe_subscriptions_digest/tests/src/Functional/UserDigestTest.php
```diff
@@ -0,0 +1,191 @@
...
+      'label' => 'Bi-weekly',
+      'interval' => '2 weeks',
+    ])->save();
+    $this->drupalGet($path);
+    $this->assertEquals([
+      'Send immediately' => 'Send immediately',
+      'message_digest:daily' => 'Daily',
+      'message_digest:weekly' => 'Weekly',
+      'message_digest:bi_weekly' => 'Bi-weekly',
+
```
brummbar: Nit: empty line.

brummbar (changes requested):

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
AaronGilMartinez: Added

modules/oe_subscriptions_digest/tests/src/Functional/UserDigestTest.php
AaronGilMartinez: Removed

modules/oe_subscriptions_digest/tests/src/Kernel/MailFlagtest.php
AaronGilMartinez: Changed

modules/oe_subscriptions_digest/tests/src/Functional/UserDigestTest.php
AaronGilMartinez: Done with second

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
AaronGilMartinez: Added the static method and removed the property

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Changed

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Moved

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,80 @@
...
+    ->set('label', $mail_prefix . ' - ' . $entity->label())
+    ->save();
+}
+
+/**
+ * Implements hook_flag_delete().
+ */
+function oe_subscriptions_digest_flag_delete(EntityInterface $entity) {
+  $mail_prefix = \Drupal::config('message_subscribe_email.settings')->get('flag_prefix');
+  if (empty($mail_prefix) || !str_starts_with($entity->id(), 'subscribe_')) {
```
brummbar: I realised now that both the flag prefixes come from a config. Let's use that instead of hardcoding.

brummbar: In tests no, we can expect that to be the default value.

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,80 @@
...
+function oe_subscriptions_digest_user_update(EntityInterface $entity) {
+  if (!$entity->hasField('message_digest') ||
+    $entity->get('message_digest')->value === $entity->original->get('message_digest')->value) {
+    return;
+  }
+  $flagging_storage = \Drupal::service('entity_type.manager')->getStorage('flagging');
+  $query = $flagging_storage->getQuery();
+  $query
+    ->accessCheck(FALSE)
+    ->condition('flag_id', 'subscribe_', 'STARTS_WITH')
```
brummbar: The message digest is stored on email flags, not the subscribe one.

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Implements hook_user_update().
+ */
+function oe_subscriptions_digest_user_update(EntityInterface $entity) {
+  if (!$entity->hasField('message_digest') ||
+    $entity->get('message_digest')->value === $entity->original->get('message_digest')->value) {
+    return;
+  }
```
brummbar:
```suggestion
  if (
    !$entity->hasField('message_digest') ||
    $entity->get('message_digest')->value === $entity->original->get('message_digest')->value
  ) {
    return;
  }
```

modules/oe_subscriptions_digest/tests/src/Functional/UserDigestTestBase.php
```diff
@@ -0,0 +1,179 @@
...
+      'interval' => '1 month',
+    ])->save();
+    $this->drupalGet($path);
+    $this->assertEquals([
+      'Send immediately' => 'Send immediately',
+      'message_digest:daily' => 'Daily',
+      'message_digest:weekly' => 'Weekly',
+      'message_digest:bi_weekly' => 'Bi-weekly',
+      'message_digest:monthly' => 'Monthly',
+    ], $this->getOptions($select));
```
brummbar, line 84: This stuff goes in setup method.

modules/oe_subscriptions_digest/tests/src/Functional/UserDigestTestBase.php
```diff
@@ -0,0 +1,179 @@
...
+    ])->save();
+    FieldConfig::create([
+      'field_name' => 'message_digest',
+      'entity_type' => 'flagging',
+      'label' => 'ABC',
+      'bundle' => 'another_flag',
+      'description' => '',
+      'required' => FALSE,
+      'settings' => [],
+    ])->save();
```
brummbar: Not needed.

brummbar (changes requested):

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Changed to config

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Changed

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Changed

modules/oe_subscriptions_digest/tests/src/Functional/UserDigestTestBase.php
AaronGilMartinez: Do you want to move the intervals creation? I wanted to check that after adding new digest options they appear.

modules/oe_subscriptions_digest/tests/src/Functional/UserDigestTestBase.php
AaronGilMartinez: Removed. For default flags provided by message_digest_ui we have the frequency field, but the duplicated with the email prefix we don't, so I added a field creation on those in the same hook.

AaronGilMartinez: Missunderstood on my side, changing

modules/oe_subscriptions_digest/tests/src/Functional/UserDigestTestBase.php
brummbar: You are right, don't move them, it was just the field config to move.

modules/oe_subscriptions_digest/tests/src/Functional/AnonymousUserDigestTest.php
```diff
@@ -0,0 +1,51 @@
...
+    $user_one = DecoupledAuthUser::create([
+      'mail' => $this->randomMachineName() . '@example.com',
+      'name' => NULL,
+      'status' => 1,
+      'roles' => ['anonymous_subscriber'],
+    ]);
+    $user_one->save();
+    $this->doTestDigestPreferences($user_one, $fn_get_path);
+
+    $user_two = DecoupledAuthUser::create([
```
brummbar: I changed my mind on this, create a new test method. It will add installation time but it's the correct way.

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,88 @@
...
+  $flaggings = $flagging_storage->loadMultiple($query->execute());
+  foreach ($flaggings as $flagging) {
+    if ($flagging->hasField('message_digest')) {
+      $flagging->set('message_digest', $entity->get('message_digest')->value)->save();
+    }
+  }
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
```
brummbar:
```suggestion
 * Implements hook_form_FORM_ID_alter() for oe_subscriptions_user_subscriptions_form.
```

brummbar (changes requested):

modules/oe_subscriptions_digest/tests/src/Functional/AnonymousUserDigestTest.php
AaronGilMartinez: Splited, both

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Added

modules/oe_subscriptions_digest/src/FlagHelper.php
```diff
@@ -0,0 +1,80 @@
...
+
+use Drupal\flag\FlaggingInterface;
+use Drupal\flag\FlagInterface;
+
+/**
+ * Class to manage digest flag.
+ *
+ * @internal
+ */
+class FlagHelper {
```
AaronGilMartinez: @todo Should this be a service?

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
```diff
@@ -0,0 +1,89 @@
...
+ * Implements hook_user_update().
+ */
+function oe_subscriptions_digest_user_update(EntityInterface $entity) {
+  if (
+    !$entity->hasField('message_digest') ||
+    $entity->get('message_digest')->value === $entity->original->get('message_digest')->value
+  ) {
+    return;
+  }
+  $mail_prefix = \Drupal::config('message_subscribe_email.settings')->get('flag_prefix');
```
brummbar: Can't we have a ::getFlagPrefix() method at this point?

brummbar (changes requested):

modules/oe_subscriptions_digest/oe_subscriptions_digest.module
AaronGilMartinez: Sure, added the shortcut. It could be shorter, but wanted to keep some flexibility to call one or another.

brummbar (approved):

PR #16, opened 2024-01-11
OEL-2786: Use a page reload instead of a ajax command to display messages.

AaronGilMartinez:


modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -200,17 +195,11 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
    * {@inheritdoc}
    */
   protected function successfulAjaxSubmit(array $form, FormStateInterface $form_state) {
+    $form_state->disableRedirect(FALSE);
```
brummbar, line 204: At first this scared me, as core explicitly disables this. But I understand you are doing this to retrieve the redirect set in the standard submit callback.
Let's:

1. add a comment that this is our intention: so developers that don't know that redirects are disabled in ajax requests (like myself) don't need to check up the code.
2. Store the current status of redirect (`isRedirectDisabled`) in a variable.
3. Collect the redirect from the form state.
4. Restore the original value.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -115,6 +123,18 @@ public function testModalForm(): void {
     $this->assertSubscriptionCreateMailStatusMessage();
     $this->assertCount(1, $this->getMails());
     $this->assertMail('to', 'test@test.com');
+    $assert_session->addressEquals($node->toUrl());
+
+    // Test confirm message override.
+    $this->config('system.theme')->set('default', 'oe_subscriptions_anonymous_theme_test')->save();
```
brummbar, line 132: I appreciate a lot the effort on adding the tests, but I am not sure about this. It does end up checking that the page has been reloaded rather than using a command to render the messages...
Let's keep it because I like the idea of testing it, but let's add a big comment that explains the reason why this is there: the message command doesn't take into account template overrides etc etc.

modules/oe_subscriptions_anonymous/tests/themes/oe_subscriptions_anonymous_theme_test/oe_subscriptions_anonymous_theme_test.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'OE Subscriptions anonymous theme test'
+type: theme
+base theme: stark
+description: 'Support theme for anonymous subscriptions tests.'
+package: Testing
```
brummbar, line 5: Add `hidden: true`

brummbar (changes requested):

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Done

modules/oe_subscriptions_anonymous/tests/themes/oe_subscriptions_anonymous_theme_test/oe_subscriptions_anonymous_theme_test.info.yml
AaronGilMartinez: Added

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -194,23 +189,25 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
    * {@inheritdoc}
    */
   protected function successfulAjaxSubmit(array $form, FormStateInterface $form_state) {
+    // Adding redirection from regular form submit variables. By default,
+    // Drupal\Core\Form\FormBuilder disables the redirection in AJAX submission.
+    // This is byassed in order to allow the message template from theme take
+    // place. Drupal\Core\Ajax\MessageCommand does not use theme template.
+    $form_state->disableRedirect($form_state->get('redirect_disabled'));
     $response = new AjaxResponse();
+    $url = new RedirectCommand($form_state->getRedirect()->toString());
```
brummbar: This is not really what I meant with "store the value". First, we should never use the form build array to store data like that. In general, what I meant was:
```
$is_redirect_disabled = $form_state->isRedirectDisabled();
$form_state-> disableRedirect(FALSE);
$redirect = $form_state->getRedirect();
$form_state->disableRedirect($is_redirect_disabled);
```

Regarding the comment, I wanted a comment on why we are forced to disable the redirect. The comment about replacing the ajax message with a reload is already in the test, here we need to explain why we cannot retrieve the redirect from the form state without changing the form state value.

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
```diff
@@ -115,6 +123,22 @@ public function testModalForm(): void {
     $this->assertSubscriptionCreateMailStatusMessage();
     $this->assertCount(1, $this->getMails());
     $this->assertMail('to', 'test@test.com');
+    $assert_session->addressEquals($node->toUrl());
+
+    // Test confirm message override.
+    // Drupal\Core\Ajax\MessageCommand does not use theme template, and
+    // overrides are not taken in place. In order to solve the issue, current
+    // implementation adds a redirection in AjaxSubmit with values from regular
+    // submit, adding the confirm message through the messenger service.
```
brummbar: Close, but not enough. Let's rephrase:
```
Test that the modal reloads the page in order to show the success message, rather then using
the message AJAX command. The message command handles the status message rendering directly, without
taking into account template overrides.

brummbar (changes requested):

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
AaronGilMartinez: Fixed and changed

modules/oe_subscriptions_anonymous/tests/src/FunctionalJavascript/SubscribeModalFormTest.php
AaronGilMartinez: Changed

brummbar (approved):

PR #18, opened 2024-01-16
OEL-2788: Add a template for the anonymous subscribe link.

AaronGilMartinez:


modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.module
```diff
@@ -58,5 +59,27 @@ function oe_subscriptions_anonymous_theme($existing, $type, $theme, $path) {
...
+        'attributes' => [],
+      ],
+    ],
   ];
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_subscriptions_anonymous_preprocess_oe_subscriptions_anonymous_link(&$variables) {
```
brummbar: The module that declares the theme hook should create template_hook().

modules/oe_subscriptions_anonymous/tests/src/Functional/TemplatesTest.php
```diff
@@ -0,0 +1,53 @@
...
+   * {@inheritdoc}
+   */
+  protected $defaultTheme = 'stark';
+
+  /**
+   * Tests the anonymous subscribe link template.
+   */
+  public function testLinkTemplate(): void {
+    $assert_session = $this->assertSession();
+    $this->drupalGet('/oe-subscriptions-anonymous-test/template-link-test');
```
brummbar: Use a kernel test.

brummbar (changes requested):
Move the button classes to the template declaration too.

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
```diff
@@ -0,0 +1,95 @@
...
+      '#url' => $front_url,
+    ]);
+
+    // Test link with attributes.
+    $this->assertAnonymousLink([
+      '#title' => 'Link with attributtes',
+      '#url' => $front_url,
+      '#attributes' => [
+        'class' => ['class-1', 'class-2'],
+        'attribute-2' => 'attribute-2-value',
```
brummbar: Use target => '_blank'

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
```diff
@@ -0,0 +1,95 @@
...
+    ]);
+  }
+
+  /**
+   * Asserts that an anonymous link template is rendered with the given values.
+   *
+   * @param array $variables
+   *   The expected values.
+   */
+  protected function assertAnonymousLink(array $variables): void {
```
brummbar: Very good! Now, when we run assertions I like to run them by comparing the expected result directly in the easiest form possible (when we can). In this case, we can pass the expected route as string, text as string, attributes as array.

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
```diff
@@ -0,0 +1,95 @@
...
+          $v = implode(' ', $v);
+        }
+        if ($k === 'class') {
+          // For the classes we check that contains the specified elements given
+          // we have some predefined classes.
+          $this->assertStringContainsString($v, $link->attr($k));
+          continue;
+        }
+        $this->assertEquals($link->attr($k), $v);
+      }
```
brummbar: This explains my comment above. You need to replicate the code of the attributes processing done by core in order to make assertions. If instead you passed an array of attributes, the code would be as simple as:
```
// Retrieve all attributes present in the node.
$attributes = array_map(static fn ($attr) => $attr->value, iterator_to_array($link->getNode(0)->attributes));
$this->assertEquals($expected_attributes, $attributes);
```

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
```diff
@@ -0,0 +1,95 @@
...
+    $values = ['#theme' => 'oe_subscriptions_anonymous_link'] + $variables;
+    $html = (string) $this->container->get('renderer')->renderRoot($values);
+    $crawler = new Crawler($html);
+    $link = $crawler->filter('a');
+
+    if (!empty($variables['#title'])) {
+      $this->assertEquals($link->text(), $variables['#title']);
+    }
+
+    if (!empty($variables['#url']) && $variables['#url'] instanceof Url) {
```
brummbar: If url is not passed, we should have tested that href is empty. Anyway we will do some changes in the template so this will change.

modules/oe_subscriptions_anonymous/templates/oe-subscriptions-anonymous-link.html.twig
```diff
@@ -0,0 +1,14 @@
...
+ * Default theme implementation for anonymous subscribe link.
+ * Available variables:
+ * - link: The link text.
+ * - url: The link destination.
+ * - attributes: associative array containing the properties of the element.
+ */
+#}
+{% apply spaceless %}
+  {# Set set URL from variables #}
+  <a{{ attributes.addClass(['button', 'button--small']) }}>{{ title }}</a>
```
brummbar: You did everything correct, it's me that did a mistake. I got fooled by flag approach, which is a bit D7 in this regard.
Our template should use the core function link() and be
```
{{ link(title, url, attributes.addClass(['button', 'button--small']) }}
```
This allows to completely drop the template_preprocess function. It will also force people to put a url 😄

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
```diff
@@ -0,0 +1,95 @@
...
+          // For the classes we check that contains the specified elements given
+          // we have some predefined classes.
+          $this->assertStringContainsString($v, $link->attr($k));
+          continue;
+        }
+        $this->assertEquals($link->attr($k), $v);
+      }
+    }
+    // Button classes are always present.
+    $this->assertStringContainsString('button button--small', $link->attr('class'));
```
brummbar: Yet another mistake by me: default templates should have less visual opinion on components. Let's drop the classes completely: it will show like a normal link.

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Tests the templates.
+ */
+class TemplatesTest extends KernelTestBase {
+
+  /**
+   * Tests the anonymous subscribe link template.
+   */
+  public function testLinkTemplate(): void {
+    $front_url = Url::fromRoute('<front>');
```
brummbar: Let's also try with a string url (now that we change how the thing works).

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
```diff
@@ -0,0 +1,95 @@
...
+
+namespace Drupal\Tests\oe_subscriptions_anonymous\Kernel;
+
+use Drupal\Core\Url;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Tests the templates.
+ */
+class TemplatesTest extends KernelTestBase {
```
brummbar: I'm afraid we might mix a bit twig templates and mail templates. Maybe we can rename this, maybe TwigTemplatesTest?

brummbar (changes requested):

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
AaronGilMartinez: Added

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
AaronGilMartinez: Done, better than getting keys and values

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
AaronGilMartinez: Replaced

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
AaronGilMartinez: ok

modules/oe_subscriptions_anonymous/templates/oe-subscriptions-anonymous-link.html.twig
AaronGilMartinez: Removed preprocess and added function

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
AaronGilMartinez: Removed and restored in extra field

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
AaronGilMartinez: Added case

modules/oe_subscriptions_anonymous/tests/src/Kernel/TemplatesTest.php
AaronGilMartinez: Renamed

brummbar (approved):

PR #20, opened 2024-01-19
OEL-2794: Update components to run on Drupal 10.2.x

AaronGilMartinez:


composer.json
```diff
@@ -8,7 +8,7 @@
     "require": {
         "php": ">=8.1",
         "cweagans/composer-patches": "^1.7",
-        "drupal/core": "^9.5 || ^10",
+        "drupal/core": "^10.1 || ^10.2",
```
brummbar: Let's just put ^10 here, without further restrictions.

composer.json
```diff
@@ -18,10 +18,10 @@
     "require-dev": {
         "composer/installers": "^2.0",
         "drupal/config_devel": "^1.9",
-        "drupal/core-composer-scaffold": "^9.4 || ^10",
-        "drupal/core-dev": "^9.4 || ^10",
+        "drupal/core-composer-scaffold": "^10.1 || ^10.2",
```
brummbar: Same here.

brummbar (changes requested):

composer.json
AaronGilMartinez: Done

composer.json
AaronGilMartinez: Done

brummbar (approved):

PR #23, opened 2024-01-25
OEL-2789: Add route subscriber to modify message_subscribe.

AaronGilMartinez:


src/Routing/SubscriptionsRouteSubscriber.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+class SubscriptionsRouteSubscriber extends RouteSubscriberBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterRoutes(RouteCollection $collection) {
+    $message_subscribe_routes = [
+      // Module: message_subscribe.
+      'message_subscribe.admin_settings',
```
brummbar: This should stay, only the user routes are duplicated for us.

brummbar (changes requested):

src/Routing/SubscriptionsRouteSubscriber.php
AaronGilMartinez: Removed

brummbar (approved):

PR #26, opened 2024-02-29
OEL-2842: Order elements in subscriptions page form.

AaronGilMartinez:


modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
```diff
@@ -60,12 +60,20 @@ public function subscriptionsFormAlter(array &$form, FormStateInterface $form_st
     $message_digest = $user->get('message_digest');
     $storage_definition = $message_digest->getFieldDefinition()->getFieldStorageDefinition();
 
+    // Move down all render elements except language preference.
+    array_walk($form, function (&$v, $k) {
```
brummbar: This is over-engineered. We own the form, we can declare weights suitable for us (0,10,20) or even use decimals in weights such as 0.001. But let's just put 0, 10, 20 to the elements in the form and here we put a 5.

brummbar (changes requested):

modules/oe_subscriptions_digest/src/SubscriptionsFormAlter.php
AaronGilMartinez: Understood

brummbar (approved):

PR #27, opened 2024-02-29
OEL-2806: Use the same warning message for user subscriptions request token email.

AaronGilMartinez:
…t message.

### DIFF
--- a/modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.module
+++ b/modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.module
@@ -62,8 +62,8 @@ function oe_subscriptions_anonymous_theme($existing, $type, $theme, $path) {
   return [
     'oe_subscriptions_anonymous_status_message' => [
       'variables' => [
-        'title' => '',
-        'message' => '',
+        'title' => NULL,
+        'message' => NULL,
       ],
     ],
     'oe_subscriptions_anonymous_link' => [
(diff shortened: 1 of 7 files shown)


modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.module
```diff
@@ -61,7 +61,10 @@ function oe_subscriptions_anonymous_mail($key, &$message, $params) {
 function oe_subscriptions_anonymous_theme($existing, $type, $theme, $path) {
   return [
     'oe_subscriptions_anonymous_message_confirm' => [
-      'variables' => [],
+      'variables' => [
```
brummbar, line 64: The work is good, but I would:

1. Not default any values, require title/message to be passed always.
2. Change the template hook name: maybe we could name it `oe_subscriptions_anonymous_status_message` ?

brummbar (changes requested):

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.module
AaronGilMartinez: The name change is Ok. 
About the require of the title/message, removed default values, and to be sure, you want to make them mandatory? Like no displaying anything or throwing an error?

brummbar: Often Drupal uses NULL and doesn't really check anything. Let's check for each value and do not print anything if not there.

AaronGilMartinez: Changed to NULL, added checks

brummbar (approved):

PR #31, opened 2024-03-15
OEL-2853: Add support for HTML emails

AaronGilMartinez:


modules/oe_subscriptions_anonymous/src/MailTemplate/UserSubscriptionsAccess.php
```diff
@@ -36,13 +36,23 @@ public static function create(ContainerInterface $container) {
...
    */
-  public function prepare(array &$message, array $params): void {
+  public function prepare(array $params, bool $has_html = FALSE): array {
     $mail = $params['email'];
     $hash = $this->tokenManager->get($mail, 'user_subscriptions_page');
     $site_url = Url::fromRoute('<front>', [], ['absolute' => TRUE])->toString();
+    $message = [];
 
     $variables = [
       '@site_url' => $site_url,
```
drishu: Let's make this a link now that it's html

modules/oe_subscriptions_anonymous/src/MailTemplate/MailTemplateHelper.php
```diff
@@ -0,0 +1,37 @@
...
+   */
+  public static function getKeyClass(string $key): string {
+    return match ($key) {
+      'subscription_create' => SubscriptionCreate::class,
+      'user_subscriptions_access' => UserSubscriptionsAccess::class,
+    };
+  }
+
+  /**
+   * Instantiates an object of the given mail template key.
```
drishu: Gets..

modules/oe_subscriptions_anonymous/src/MailTemplate/MailTemplateInterface.php
```diff
@@ -12,11 +12,22 @@ interface MailTemplateInterface {
...
+   * @param bool $has_html
+   *   If the mail has HTML.
+   *
+   * @return array
+   *   The processed subject and body.
+   */
+  public function prepare(array $params, bool $has_html): array;
+
+  /**
+   * Retrieves parameter keys used in the mail template.
```
drishu: Gets..

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
```diff
@@ -0,0 +1,68 @@
...
+use Drupal\symfony_mailer\Address;
+use Drupal\symfony_mailer\EmailFactoryInterface;
+use Drupal\symfony_mailer\EmailInterface;
+use Drupal\symfony_mailer\Processor\EmailBuilderBase;
+
+/**
+ * Defines the Email Builder plug-in for oe_subscriptions_anonymous module.
+ *
+ * @EmailBuilder(
+ *   id = "oe_subscriptions_anonymous",
```
drishu, line 18: the id is a bit confusing, can we use something more descriptive

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
```diff
@@ -0,0 +1,68 @@
...
+ *   sub_types = {
+ *     "subscription_create" = @Translation("Subscription creation"),
+ *     "user_subscriptions_access" = @Translation("Subscriptions access request"),
+ *   },
+ * )
+ */
+class AnonymousEmailBuilder extends EmailBuilderBase {
+
+  /**
+   * Saves the parameters for a newly created email.
```
drishu: Creates..

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
```diff
@@ -0,0 +1,68 @@
...
+   */
+  public function build(EmailInterface $email): void {
+    $params = [];
+    $mail_template = MailTemplateHelper::getMailTemplate($email->getSubType());
+
+    foreach ($mail_template->getParameters() as $param) {
+      $params[$param] = $email->getParam($param);
+    }
+
+    $message = $mail_template->prepare($params, TRUE);
```
drishu: I see the email is always set to html, but in the ticket description asked for it to be optional "The module must be optional." 
If I misunderstand for lack of context please let me know.

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Saves the parameters for a newly created email.
+   *
+   * @param \Drupal\symfony_mailer\EmailInterface $email
+   *   The email to modify.
+   * @param array $params
+   *   The email parameters.
+   */
+  public function createParams(EmailInterface $email, array $params = []): void {
+    foreach (MailTemplateHelper::getKeyClass($email->getSubType())::getParameters() as $param) {
```
drishu: this is a bit confusing for me as getKeyClass returns a string, is it possible to decompose it a bit, its hard to read

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -0,0 +1,179 @@
...
+      'entity_type' => 'node',
+      'bundles' => ['article'],
+    ]);
+    $article = $this->drupalCreateNode([
+      'type' => 'article',
+      'status' => 1,
+    ]);
+    $assert_session = $this->assertSession();
+
+    // Test confim subcription HTML mail content.
```
drishu: typo

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -0,0 +1,179 @@
...
+    // Subscribe to an article.
+    $this->drupalGet($article->toUrl());
+    $this->clickLink('Subscribe');
+    $assert_session->fieldExists('Your e-mail')->setValue('test@test.com');
+    $assert_session->fieldExists('I have read and agree with the data protection terms.')->check();
+    $assert_session->buttonExists('Subscribe me')->press();
+    $this->assertSubscriptionCreateMailStatusMessage();
+
+    // Check the subscribe mail content.
```
drishu:
```suggestion
    // Check the subscription mail content.
```

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -0,0 +1,179 @@
...
+    $mail = $this->readMail();
+    $this->assertTo('test@test.com');
+    $this->assertSubject("Confirm your subscription to {$article->label()}");
+    $this->assertMailBreakLines(3, $mail);
```
drishu: the editor is telling me that the expected type is a mismatch

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -0,0 +1,179 @@
...
+    $links = $this->getMailLinks($mail);
+    $this->assertEquals($article->label(), $links[0]->textContent);
+    $this->drupalGet($links[0]->getAttribute('href'));
+    $assert_session->addressEquals($article->toUrl());
+
+    // Check confirmation link.
+    $this->assertEquals('Confirm my subscription', $links[1]->textContent);
+    $this->drupalGet($links[1]->getAttribute('href'));
+    $assert_session->statusMessageContains('Your subscription request has been confirmed.', 'status');
+    $assert_session->addressEquals($article->toUrl());
```
drishu: cast to string

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -0,0 +1,179 @@
...
+    $assert_session->elementExists('css', 'table.user-subscriptions');
+  }
+
+  /**
+   * Retrieves the DOM elements from existing links in a mail.
+   *
+   * @param \Drupal\symfony_mailer\Email $mail
+   *   The mail.
+   *
+   * @return \DOMElement[]
```
drishu: add ext-dom to composer requirements

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -0,0 +1,179 @@
...
+   *
+   * @param string[] $expected_texts
+   *   A list of expected text content to search in the mail body.
+   * @param \Drupal\symfony_mailer\Email $mail
+   *   The mail.
+   */
+  protected function assertMailContent(array $expected_texts, Email $mail): void {
+    $crawler = new Crawler($mail->getHtmlBody());
+    foreach ($expected_texts as $expected) {
+      $this->assertStringContainsString($expected, $crawler->text());
```
drishu: you may be doing getHtmlBody but then there is no actual tag being tested here

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -0,0 +1,179 @@
...
+use Drupal\symfony_mailer_test\MailerTestTrait;
+use Drupal\Tests\BrowserTestBase;
+use Drupal\Tests\flag\Traits\FlagCreateTrait;
+use Drupal\Tests\oe_subscriptions_anonymous\Trait\StatusMessageTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Tests the HTML in mails.
+ */
+class HtmlMailsTest extends BrowserTestBase {
```
drishu, line 19: I think the idea behind the test is not right at the moment, it doenst actually check html and most of what it tests is already tested in other tests. Once the refactoring for optional html email is done, then I think we can rethink the existing traits to account for this, and change the anonymous testing or even add a new test if needed.

drishu (changes requested):

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
AaronGilMartinez: Has to match module name

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
AaronGilMartinez: Splited

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
AaronGilMartinez: The EmailBuilder will add HTML mails in case the symfony_mailer is enabled.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: fixed

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: Changed

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: added

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: Me too, the trait provided by symfony_mailer to read mails its using a property to store the last mail with type \Drupal\symfony_mailer\EmailInterface but declaring a different type \Symfony\Component\Mime\Email in the function docs for the return.

You can check in oe_subscriptions\build\modules\contrib\symfony_mailer\tests\modules\symfony_mailer_test\src\MailerTestTrait.php

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: It seems the correct method but it's used in another phase of the mail building, can't be used to check the final result only for pre-render. You can change the method and test will throw an error about phases.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: The support for HTML emails is based on symfony_mailer, and the presence of the module is what makes the HTML mails optional.

Not sure if we should move all this logic to the traits, meaning that we would have to check if the module is enabled there, and also manage this option by enabling/disabling the module for the tests that contains mails.

The idea was to create a test to cover all the possible mails and HTML for each.

modules/oe_subscriptions_anonymous/src/MailTemplate/UserSubscriptionsAccess.php
AaronGilMartinez: Changed to a link

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: Clarifications call: move assertion to a method to check the HTML output with placeholders with expected values for the mails.

AaronGilMartinez: After checking, output HTML can't be completely asserted because of links tokens, some of the content of the mail contains links with tokens that can be accessed only through the mail, so the only way to check them is using them.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.module
```diff
@@ -48,11 +47,10 @@ function _oe_subscriptions_anonymous_flush_extra_field(EntityInterface $entity)
  * Implements hook_mail().
  */
 function oe_subscriptions_anonymous_mail($key, &$message, $params) {
```
AaronGilMartinez: Added types in all hooks

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
drishu: not quite please fix also "subcription"

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
drishu: a yes, I didnt see that in oe_subscriptions_anonymous_mail the html param is not set

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
drishu: true

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
drishu: I dont see the change

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
drishu: No, if I print out $mail->getHtmlBody() I get a bunch of html that you can assert against

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
drishu: Lets try to get the hash similar to what was done in SubscriptionRequestRoutesTest, give it a try

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: true

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: It was there indeed I missed the change, with new version just check message or title of the page, but can add it again.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: I was refering to getBody(), I missunderstood

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: I've adapted the test, now we have one function to check each mail type and I check that the HTML is present and where, about links with tokens couldn't use the manager to get the URL, the service generates a new token every time you resquest it. I check the link performing the click instead.

modules/oe_subscriptions_anonymous/src/MailTemplate/MailTemplateHelper.php
```diff
@@ -0,0 +1,37 @@
...
+   *
+   * @return string
+   *   The associated class.
+   */
+  public static function getKeyClass(string $key): string {
+    return match ($key) {
+      'subscription_create' => SubscriptionCreate::class,
+      'user_subscriptions_access' => UserSubscriptionsAccess::class,
+    };
+  }
```
brummbar: Let's remove this, just use the getMailTemplate().

modules/oe_subscriptions_anonymous/src/MailTemplate/MailTemplateInterface.php
```diff
@@ -12,11 +12,22 @@ interface MailTemplateInterface {
...
-   *   Message parts.
    * @param array $params
    *   Mail parameters.
+   * @param bool $has_html
+   *   If the mail has HTML.
+   *
+   * @return array
+   *   The processed subject and body.
+   */
+  public function prepare(array $params, bool $has_html = FALSE): array;
```
brummbar:
```suggestion
  public function prepare(array $params, bool $is_html = FALSE): array;
```

modules/oe_subscriptions_anonymous/src/MailTemplate/SubscriptionCreate.php
```diff
@@ -81,15 +93,17 @@ public function prepare(array &$message, array $params): void {
...
+Click the following link to confirm your subscription: @confirm_link<br>
+If you no longer wish to subscribe, click on the link bellow: @cancel_link<br>
+If you didn't subscribe to these updates or you're not sure why you received this e-mail, you can delete it.
 You will not be subscribed if you don't click on the confirmation link above.", $variables);
-    $message['subject'] .= $this->t('Confirm your subscription to @label', [
+    $message['subject'] = $this->t('Confirm your subscription to @label', [
       '@label' => $entity->label(),
     ]);
-    $message['body'][] = MailFormatHelper::htmlToText($body);
+    $message['body'] = $has_html ? $body : MailFormatHelper::htmlToText($body);
```
brummbar: Should we just move the is_html and conversion to the hook_mail implementation?

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
```diff
@@ -0,0 +1,70 @@
...
+   * @param \Drupal\symfony_mailer\EmailInterface $email
+   *   The email to modify.
+   * @param array $params
+   *   The email parameters.
+   */
+  public function createParams(EmailInterface $email, array $params = []): void {
+    $mail_template = MailTemplateHelper::getKeyClass($email->getSubType());
+
+    foreach ($mail_template::getParameters() as $param) {
+      assert($params[$param] != NULL);
```
brummbar: Let's remove this, it works only in dev mode. If we need to run checks, we should throw errors. But there's no need.

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build(EmailInterface $email): void {
+    $params = [];
+    $mail_template = MailTemplateHelper::getMailTemplate($email->getSubType());
+
+    foreach ($mail_template->getParameters() as $param) {
+      $params[$param] = $email->getParam($param);
```
brummbar, line 63: We need to have these defined as variables so that they can be used in template overrides.
We also need to define the mailer_override configurations so that people can easily override the text.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -0,0 +1,250 @@
...
+      $link = $crawler->filterXPath("//a[contains(text(),'{$expected_link['text']}')]");
+      $this->assertCount(1, $link);
+      if (isset($expected_link['url'])) {
+        $this->assertEquals($expected_link['url'], $link->attr('href'));
+        continue;
+      }
+      $skipped_urls[] = $link->attr('href');
+    }
+
+    return $skipped_urls;
```
brummbar: Can't we just call this "text_only_urls"? Urls of which we checked only the text.

brummbar (changes requested):

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: Renamed

modules/oe_subscriptions_anonymous/src/MailTemplate/MailTemplateHelper.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/src/MailTemplate/MailTemplateInterface.php
AaronGilMartinez: Noted, removing parameter since we want to have only message parameters and handle this in the hook

modules/oe_subscriptions_anonymous/src/MailTemplate/SubscriptionCreate.php
AaronGilMartinez: Yes, makes sense

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
AaronGilMartinez: Removed

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
AaronGilMartinez: Added variables and configurations to override

AaronGilMartinez (commented):
Approved

drishu (approved):

brummbar (approved):

PR #34, opened 2024-05-08
DARWIN-1045: Provide link to user profile and/or subscriptions page

AaronGilMartinez:


modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.tokens.inc
```diff
@@ -0,0 +1,78 @@
...
+ * Implements hook_tokens().
+ */
+function oe_subscriptions_anonymous_tokens(string $type, array $tokens, array $data, array $options, BubbleableMetadata $bubbleable_metadata) {
+  $replacements = [];
+
+  if (
+    $type == 'oe_subscriptions_anonymous' &&
+    isset($tokens['subscriptions_page']) &&
+    !empty($data['message'])
+  ) {
```
donquixote: I like this formatting better:
```suggestion
  if ($type === 'oe_subscriptions_anonymous'
    && isset($tokens['subscriptions_page'])
    && !empty($data['message'])
  ) {
```
Actually, Drupal core is inconsistent about this.
And in /vendor/ you also find different formats.
See
https://www.drupal.org/project/coding_standards/issues/3392321
https://www.drupal.org/project/coding_standards/issues/1539712
Look for "&&" to find relevant comments.

Ultimately you can decide this with the OEL team.

modules/oe_subscriptions_anonymous/tests/src/Kernel/TwigTemplatesTest.php
```diff
@@ -50,7 +63,8 @@ public function testLinkTemplate(): void {
       '',
       [
         'href' => $front_url->toString(),
-      ]
+      ],
+      $template_name
```
donquixote: Multi line argument list should have comma after the last item.
I thought this would be in automatic code style checks but seems it is not?

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.tokens.inc
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ * Implements hook_tokens().
+ */
+function oe_subscriptions_anonymous_tokens(string $type, array $tokens, array $data, array $options, BubbleableMetadata $bubbleable_metadata) {
+  $replacements = [];
+
+  if (
+    $type == 'oe_subscriptions_anonymous' &&
```
brummbar:
```suggestion
    $type === 'oe_subscriptions_anonymous' &&
```

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.module
```diff
@@ -71,6 +71,13 @@ function oe_subscriptions_anonymous_theme(array $existing, string $type, string
         'attributes' => [],
       ],
     ],
+    'oe_subscriptions_anonymous_subscriptions_link' => [
```
brummbar: We shouldn't add templates for all the links, we did it for the flag because it's a special case. Anyway see below why this is not needed.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.tokens.inc
```diff
@@ -0,0 +1,78 @@
...
+declare(strict_types=1);
+
+use Drupal\Core\Render\BubbleableMetadata;
+use Drupal\Core\Url;
+
+/**
+ * Implements hook_token_info().
+ */
+function oe_subscriptions_anonymous_token_info(): array {
+  $type = [
```
brummbar: All the token approach is wrong.
Tokens works off some data, eg `node:field_something`, `user:cancel-url`, and they are recursively generated. E.g. given there's a message as context, and the message has a user property, you get `[message:author:cancel-url]`. 
Below we check if there's message as data, and in that case we add the token in the type.
The correct solution was to just add the `subscriptions-page-url` (note the dashes and not underscores) to the user type.
Furthermore, the token shouldn't handle a link, but just the url (as above). It might be needed in a plain text email, so there's no html available, or it might need further customisation in an HTML context.

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.tokens.inc
```diff
@@ -0,0 +1,78 @@
...
+    !empty($data['message'])
+  ) {
+    /** @var \Drupal\message\Entity\Message $message */
+    $message = $data['message'];
+    $route = 'oe_subscriptions_anonymous.user_subscriptions.request_access';
+    $options = [
+      'absolute' => TRUE,
+    ];
+
+    if ($message->getOwner() !== NULL && $message->getOwner()->isCoupled()) {
```
brummbar: We are adding a polymorphic token that works for both type of users. But if a developer enables only the main oe_subscriptions module, no token is added. The main module should provide the token that gives access to the `oe_subscriptions.user.subscriptions` route, and the anonymous module should change said token for decoupled users.

brummbar (changes requested):

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.module
AaronGilMartinez: Ok, removing

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.tokens.inc
AaronGilMartinez: Understood, I was applying wrong logic, thanks

modules/oe_subscriptions_anonymous/oe_subscriptions_anonymous.tokens.inc
AaronGilMartinez: Moved to main module, and anonymous alters the token

oe_subscriptions.tokens.inc
```diff
@@ -0,0 +1,62 @@
...
+    // Adding the URL to the subscriptions page will lead into an 403 for
+    // non-logged users, we add a login URL with the subscriptions page
+    // destination.
+    $url = Url::fromRoute('user.login', [], [
+      'query' => [
+        'destination' => Url::fromRoute('oe_subscriptions.user.subscriptions', [
+          'user' => $data['user']->id(),
+        ])->toString(),
+      ],
+    ]);
```
brummbar: This looked good at a first glance, but now testing the prefixed token generation, I've noticed that using a token like `subscriptions-page:path` will output `/user/login` which is incorrect.
After talking with @AaronGilMartinez , we agreed that preventing the user to hit the 403 is not the scope of the token in this module. It can be either customised in the project, or a module like https://www.drupal.org/project/r4032login can be used to configure that specific url to redirect to the login page.

brummbar (changes requested):

oe_subscriptions.tokens.inc
AaronGilMartinez: Added changes to replace route

brummbar (approved):

PR #37, opened 2024-07-24
OEL-2903: [*_anonymous] Handle conflict with email already used.

donquixote:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-2903

For `oe_subscriptions_anonymous`.

### Background
If you subscribe as anonymous for the first time, we create a "decoupled" account with that email address.
If the email address is already taken, for a regular user account, this won't work.

### Problem
Currently this kind of conflict results in an error page.

### Proposed change
- When a subscription is requested, and we detect the email conflict, we send an alternative email, asking to subscribe as registered user.
- When a user visits a confirm link, and we detect the email conflict, we show a meaningful message instead of an error page. (This case only occurs if the user creates an account before they click the confirm link).

### Unrelated changes
The PR also contains some minor refactoring, TBD.

### Remaining tasks
We also want to send a different email if the user is already subscribed.
But for now we don't have a text for it.

### DIFF
--- a/modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
+++ b/modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
@@ -22,7 +22,10 @@ class AnonymousSubscriptionManager implements AnonymousSubscriptionManagerInterf
    * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entityTypeManager
    *   The entity type manager.
    */
-  public function __construct(protected FlagServiceInterface $flagService, protected EntityTypeManagerInterface $entityTypeManager) {}
+  public function __construct(
+    protected readonly FlagServiceInterface $flagService,
+    protected readonly EntityTypeManagerInterface $entityTypeManager,
+  ) {}
 
   /**
    * {@inheritdoc}
--- a/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
+++ b/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
@@ -96,9 +96,10 @@ class SubscriptionAnonymousController extends ControllerBase {
       return $response;
     }
 
-    $this->subscriptionManager->subscribe($email, $flag, $entity_id);
     // The token has been used, so we need to invalidate it.
     $this->tokenManager->delete($email, $scope);
+
+    $this->subscriptionManager->subscribe($email, $flag, $entity_id);
     // Success message and redirection to entity.
     $this->messenger()->addMessage($this->t('Your subscription request has been confirmed.'));
 
--- a/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
+++ b/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
@@ -12,6 +12,7 @@ use Drupal\flag\FlagInterface;
 use Drupal\flag\FlagServiceInterface;
 use Drupal\oe_subscriptions\Form\UserSubscriptionsForm;
 use Drupal\oe_subscriptions_anonymous\AnonymousSubscriptionManager;
+use Drupal\oe_subscriptions_anonymous\Exception\RegisteredUserEmailException;
 use Drupal\oe_subscriptions_anonymous\TokenManagerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
@@ -99,7 +100,22 @@ class SubscriptionAnonymousController extends ControllerBase {
     // The token has been used, so we need to invalidate it.
     $this->tokenManager->delete($email, $scope);
 
-    $this->subscriptionManager->subscribe($email, $flag, $entity_id);
+    try {
+      $this->subscriptionManager->subscribe($email, $flag, $entity_id);
+    }
+    catch (RegisteredUserEmailException) {
+      // The email address is already taken.
+      // The user has already proven at this point that they own the email
+      // address.
+      // It is therefore ok to reveal that a user account with this email
+      // address is associated with a regular account.
+      $this->messenger()->addWarning(
+        '<p>' . $this->t('You have attempted to subscribe as anonymous, using an email address that is already associated with a regular account.') . '</p>'
+        . '<p>' . $this->t('If you still want to subscribe, you should log in and subscribe as a regular user.') . '</p>',
+      );
+      return $response;
+    }
+
     // Success message and redirection to entity.
     $this->messenger()->addMessage($this->t('Your subscription request has been confirmed.'));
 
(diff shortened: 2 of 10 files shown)


modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
```diff
@@ -96,9 +97,25 @@ public function confirmSubscriptionRequest(FlagInterface $flag, string $entity_i
...
     }
 
-    $this->subscriptionManager->subscribe($email, $flag, $entity_id);
     // The token has been used, so we need to invalidate it.
     $this->tokenManager->delete($email, $scope);
+
+    try {
+      $this->subscriptionManager->subscribe($email, $flag, $entity_id);
+    }
+    catch (RegisteredUserEmailException) {
```
donquixote, line 106: We could rethink the return behavior of the `->subscribe()` method.
Instead of an exception, there could be a return value indicating different types of success or failure.
For the first version I wanted to avoid scope creep.

modules/oe_subscriptions_anonymous/src/MailTemplate/EmailTaken.php
```diff
@@ -0,0 +1,84 @@
...
+
+    $message['subject'] = $this->t('Cannot subscribe to @entity_label', [
+      '@entity_label' => $variables['entity_label'],
+    ]);
+
+    $message['body'] = $this->t(
+      '<p>Thank you for showing interest in keeping up with the updates for <a href=":entity_url">@entity_label</a>!</p>
+<p>The email address you were using to subscribe is already associated with a regular account on this website.</p>
+<p>If you still want to subscribe to content updates for this item, you should log into the website, using your existing account, and then subscribe as a regular user.</p>
+<p>If you do not want to subscribe, you can ignore this message.</p>',
```
donquixote, line 75: Message text to be reviewed and refined.

donquixote: Also, we currently have duplication in the email text:
We have one place that defines the email content when using Drupal core mailer, and another place for symfony mailer.
I am not really happy about this, but it seems changing this would be out of scope for this issue.

modules/oe_subscriptions_anonymous/src/MailTemplate/MailTemplateInterface.php
```diff
@@ -29,7 +29,9 @@ public function getParameters(): array;
   public function getVariables(array $params): array;
 
   /**
-   * Prepares the mail template.
+   * Prepares the mail content if the core mailer is used.
+   *
+   * This method has no effect if the emails are sent with symfony_mailer.
```
donquixote, line 34: Please confirm if this is true :)

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
```diff
@@ -78,10 +79,13 @@ public function testForm(): void {
...
     $assert_session = $this->assertSession();
-    $this->drupalGet($article->toUrl());
 
-    // Click subscribe link.
+    // Visit the article node, and click subscribe.
+    $this->drupalGet($article->toUrl());
     $this->clickLink('Subscribe to this article');
+
+    // Without javascript, the subscribe form opens in a new page.
+    // With javascript, it would open in a modal.
```
donquixote, line 88: My goal here is that you can skim over the test by just reading the comments.

For the best result, I believe it is fine to have some level of redundancy, and not be afraid to state the obvious.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -154,20 +155,46 @@ public function buildForm(array $form, FormStateInterface $form_state, FlagInter
...
+        $mail_key = 'email_taken';
+        $mail_params = [
+          'entity_type' => $flag->getFlaggableEntityTypeId(),
+          'entity_id' => $entity_id,
+        ];
+      }
+    }
+
     // @todo Send a different e-mail when the user is already subscribed.
-    // @todo Send a different e-mail if the user is coupled.
```
donquixote, line 160: One `@todo` is gone, the other is still here, to be handled in a follow-up.

modules/oe_subscriptions_anonymous/src/AnonymousSubscriptionManager.php
```diff
@@ -22,7 +22,10 @@ class AnonymousSubscriptionManager implements AnonymousSubscriptionManagerInterf
    * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entityTypeManager
    *   The entity type manager.
    */
-  public function __construct(protected FlagServiceInterface $flagService, protected EntityTypeManagerInterface $entityTypeManager) {}
+  public function __construct(
+    protected readonly FlagServiceInterface $flagService,
+    protected readonly EntityTypeManagerInterface $entityTypeManager,
+  ) {}
```
donquixote, line 28: As a general note:
The `readonly` can cause problem in combination with `DependencySerializationTrait`, this is why we cannot use it on form classes.
But in most other places it is fine.

modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
donquixote: Also note that the proposed code change omits the `$e` variable.
The reason is that the exception message or stack does not tell us anything relevant.
We know there is exactly one reason why the exception would be thrown.

modules/oe_subscriptions_anonymous/src/MailTemplate/EmailTaken.php
donquixote: Btw the `<p>` tags are converted to line breaks in non-html email, so that between two paragraphs there is always a blank line.
Imo it is fine to have these `<p>` tags, even if they are later replaced.
TBD.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -182,86 +215,140 @@ public function testSymfonyMailerOverrides(): void {
...
+    $article_url = $this->article->toUrl()->setAbsolute()->toString();
+    $this->assertSame(
+      <<<BODY
+<p>Thank you for showing interest in keeping up with the updates for <a href="$article_url">$article_label</a>!</p>
+<p>The email address you were using to subscribe is already associated with a regular account on this website.</p>
+<p>If you still want to subscribe to content updates for this item, you should log into the website, using your existing account, and then subscribe as a regular user.</p>
+<p>If you do not want to subscribe, you can ignore this message.</p>
+BODY,
+      $this->getMailBodyWithoutWrapper('email_taken'),
+    );
```
donquixote, line 257: Imo this style of assertion is easier to read than doing separate asserts for individual tags.
It also provides more informative output on failure.

We want to avoid sensitivity to irrelevant "noise" that could be caused by external factors.
But I don't think this is a problem here.

We could also rewrite other assertions in this style, but we need to handle dynamic parts (token urls).
I created a follow-up proof-of-concept branch for this, see https://github.com/openeuropa/oe_subscriptions/compare/OEL-2903...OEL-2903-further-refactor-tests

brummbar:
PR will be re-opened under the correct branch.

PR #38, opened 2024-08-26
DARWIN-1152: Error from confirmation link in the email for corporate email

donquixote:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/DARWIN-1152

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -84,8 +84,7 @@
             "drupal/core-composer-scaffold": true,
             "php-http/discovery": false,
             "phpro/grumphp": true,
-            "phpstan/extension-installer": true,
-            "tbachert/spi": true
+            "phpstan/extension-installer": true
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -80,11 +80,12 @@
         "allow-plugins": {
             "composer/installers": true,
             "cweagans/composer-patches": true,
-            "drupal/core-composer-scaffold": true,
-            "phpro/grumphp": true,
             "dealerdirect/phpcodesniffer-composer-installer": true,
+            "drupal/core-composer-scaffold": true,
+            "php-http/discovery": false,
+            "phpro/grumphp": true,
             "phpstan/extension-installer": true,
-            "php-http/discovery": false
+            "tbachert/spi": true
         }
     }
 }
--- a/modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
+++ b/modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
@@ -3,7 +3,7 @@ status: true
 dependencies:
   module:
     - oe_subscriptions_anonymous
-id: oe_subscriptions_anonymous.email_taken
+id: oe_subscriptions_anonymous.registered_user_email_notice
 configuration:
   email_subject:
     value: 'Cannot subscribe to {{ entity_label }}'
--- a/modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
+++ b/modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
@@ -6,7 +6,7 @@ dependencies:
 id: oe_subscriptions_anonymous.registered_user_email_notice
 configuration:
   email_subject:
-    value: 'Cannot subscribe to {{ entity_label }}'
+    value: 'Please log in to subscribe to {{ entity_label }}'
   email_body:
     content:
       value: |-
--- a/modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
+++ b/modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
@@ -11,7 +11,7 @@ configuration:
     content:
       value: |-
         <p>Thank you for showing interest in keeping up with the updates for <a href="{{ entity_url }}">{{ entity_label }}</a>!</p>
-        <p>The email address you were using to subscribe is already associated with a regular account on this website.</p>
-        <p>If you still want to subscribe to content updates for this item, you should log into the website, using your existing account, and then subscribe as a regular user.</p>
-        <p>If you do not want to subscribe, you can ignore this message.</p>
+        <p>The email address you were using when trying to subscribe is already associated with a regular account on this website.</p>
+        <p>If you still want to subscribe to the updates made to the content, you can log in to the website, using your existing account, and then subscribe as a regular user.</p>
+        <p>If you do not want to subscribe or are unsure why you received this email, you can ignore this message.</p>
       format: email_html
--- a/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
+++ b/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
@@ -104,7 +104,7 @@ class SubscriptionAnonymousController extends ControllerBase {
       $this->subscriptionManager->subscribe($email, $flag, $entity_id);
     }
     catch (RegisteredUserEmailException) {
-      // The email address is already taken.
+      // The email address belongs to a registered user.
       // The user has already proven at this point that they own the email
       // address.
       // It is therefore ok to reveal that a user account with this email
--- a/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
+++ b/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
@@ -96,9 +96,10 @@ class SubscriptionAnonymousController extends ControllerBase {
       return $response;
     }
 
-    $this->subscriptionManager->subscribe($email, $flag, $entity_id);
     // The token has been used, so we need to invalidate it.
     $this->tokenManager->delete($email, $scope);
+
+    $this->subscriptionManager->subscribe($email, $flag, $entity_id);
     // Success message and redirection to entity.
     $this->messenger()->addMessage($this->t('Your subscription request has been confirmed.'));
 
--- a/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
+++ b/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Drupal\oe_subscriptions_anonymous\Controller;
 
+use Drupal\Component\Render\FormattableMarkup;
 use Drupal\Core\Controller\ControllerBase;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormBuilderInterface;
@@ -110,8 +111,20 @@ class SubscriptionAnonymousController extends ControllerBase {
       // It is therefore ok to reveal that a user account with this email
       // address is associated with a regular account.
       $this->messenger()->addWarning(
-        '<p>' . $this->t('You have attempted to subscribe as anonymous, using an email address that is already associated with a regular account.') . '</p>'
-        . '<p>' . $this->t('If you still want to subscribe to content updates for this item, you can log in to the website, using your existing account, and then subscribe as a regular user.') . '</p>',
+        // Create a wrapper to apply the '<p>' tags, so that translators don't
+        // need to add '<p>' tags in their translations.
+        // In other places this is done by rendering a theme template.
+        // But here there is not much for themes to override.
+        new FormattableMarkup(
+          // The wrapper html is not translatable, so the placeholders are
+          // completely internal.
+          '<p>@first_line</p>
+<p>@second_line</p>',
+          [
+            '@first_line' => $this->t('You have attempted to subscribe as anonymous, using an email address that is already associated with a regular account.'),
+            '@second_line' => $this->t('If you still want to subscribe to content updates for this item, you can log in to the website, using your existing account, and then subscribe as a regular user.'),
+          ],
+        ),
       );
       return $response;
     }
--- a/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
+++ b/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
@@ -111,7 +111,7 @@ class SubscriptionAnonymousController extends ControllerBase {
       // address is associated with a regular account.
       $this->messenger()->addWarning(
         '<p>' . $this->t('You have attempted to subscribe as anonymous, using an email address that is already associated with a regular account.') . '</p>'
-        . '<p>' . $this->t('If you still want to subscribe, you should log in and subscribe as a regular user.') . '</p>',
+        . '<p>' . $this->t('If you still want to subscribe to content updates for this item, you can log in to the website, using your existing account, and then subscribe as a regular user.') . '</p>',
       );
       return $response;
     }
--- a/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
+++ b/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Drupal\oe_subscriptions_anonymous\Controller;
 
-use Drupal\Component\Render\FormattableMarkup;
 use Drupal\Core\Controller\ControllerBase;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormBuilderInterface;
@@ -110,22 +109,13 @@ class SubscriptionAnonymousController extends ControllerBase {
       // address.
       // It is therefore ok to reveal that a user account with this email
       // address is associated with a regular account.
-      $this->messenger()->addWarning(
-        // Create a wrapper to apply the '<p>' tags, so that translators don't
-        // need to add '<p>' tags in their translations.
-        // In other places this is done by rendering a theme template.
-        // But here there is not much for themes to override.
-        new FormattableMarkup(
-          // The wrapper html is not translatable, so the placeholders are
-          // completely internal.
-          '<p>@first_line</p>
-<p>@second_line</p>',
-          [
-            '@first_line' => $this->t('You have attempted to subscribe as anonymous, using an email address that is already associated with a regular account.'),
-            '@second_line' => $this->t('If you still want to subscribe to content updates for this item, you can log in to the website, using your existing account, and then subscribe as a regular user.'),
-          ],
-        ),
-      );
+      $this->messenger()->addWarning($this->t(
+        // This message might be more readable with '<p>' tags. Unfortunately,
+        // in core themes, a message with multiple paragraphs looks the same as
+        // multiple separate messages. So, just use '<br>' instead.
+        'You have attempted to subscribe as anonymous, using an email address that is already associated with a regular account.<br>
+If you still want to subscribe to content updates for this item, you can log in to the website, using your existing account, and then subscribe as a regular user.',
+      ));
       return $response;
     }
 
--- a/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
+++ b/modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
@@ -12,6 +12,7 @@ use Drupal\flag\FlagInterface;
 use Drupal\flag\FlagServiceInterface;
 use Drupal\oe_subscriptions\Form\UserSubscriptionsForm;
 use Drupal\oe_subscriptions_anonymous\AnonymousSubscriptionManager;
+use Drupal\oe_subscriptions_anonymous\Exception\RegisteredUserEmailException;
 use Drupal\oe_subscriptions_anonymous\TokenManagerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
@@ -99,7 +100,22 @@ class SubscriptionAnonymousController extends ControllerBase {
     // The token has been used, so we need to invalidate it.
     $this->tokenManager->delete($email, $scope);
 
-    $this->subscriptionManager->subscribe($email, $flag, $entity_id);
+    try {
+      $this->subscriptionManager->subscribe($email, $flag, $entity_id);
+    }
+    catch (RegisteredUserEmailException) {
+      // The email address is already taken.
+      // The user has already proven at this point that they own the email
+      // address.
+      // It is therefore ok to reveal that a user account with this email
+      // address is associated with a regular account.
+      $this->messenger()->addWarning(
+        '<p>' . $this->t('You have attempted to subscribe as anonymous, using an email address that is already associated with a regular account.') . '</p>'
+        . '<p>' . $this->t('If you still want to subscribe, you should log in and subscribe as a regular user.') . '</p>',
+      );
+      return $response;
+    }
+
     // Success message and redirection to entity.
     $this->messenger()->addMessage($this->t('Your subscription request has been confirmed.'));
 
--- a/modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
+++ b/modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
@@ -17,7 +17,6 @@ use Drupal\Core\Mail\MailManagerInterface;
 use Drupal\Core\Render\RendererInterface;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\Core\Url;
-use Drupal\decoupled_auth\DecoupledAuthUserInterface;
 use Drupal\flag\FlagInterface;
 use Drupal\flag\FlagServiceInterface;
 use Drupal\oe_subscriptions_anonymous\SettingsFormAlter;
@@ -170,22 +169,20 @@ class AnonymousSubscribeForm extends FormBase {
 
     // Check if a user with this email already exists.
     // @todo Use dependency injection instead of function call.
+    // The decoupled_auth module is installed, which replaces the class for
+    // user entities.
+    /** @var \Drupal\decoupled_auth\DecoupledAuthUserInterface|null $account */
     $account = user_load_by_mail($mail);
 
-    if ($account !== FALSE) {
-      // The decoupled_auth module is installed, which replaces the class for
-      // user entities.
-      \assert($account instanceof DecoupledAuthUserInterface);
-      if ($account->isCoupled()) {
-        // The email address belongs to a regular user account, which requires
-        // regular login.
-        $mail_key = 'registered_user_email_notice';
-        $mail_params = [
-          'email' => $mail,
-          'entity_type' => $flag->getFlaggableEntityTypeId(),
-          'entity_id' => $entity_id,
-        ];
-      }
+    if ($account !== FALSE && $account->isCoupled()) {
+      // The email address belongs to a regular user account, which requires
+      // regular login.
+      $mail_key = 'registered_user_email_notice';
+      $mail_params = [
+        'email' => $mail,
+        'entity_type' => $flag->getFlaggableEntityTypeId(),
+        'entity_id' => $entity_id,
+      ];
     }
 
     // @todo Send a different e-mail when the user is already subscribed.
--- a/modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
+++ b/modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
@@ -17,6 +17,7 @@ use Drupal\Core\Mail\MailManagerInterface;
 use Drupal\Core\Render\RendererInterface;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\Core\Url;
+use Drupal\decoupled_auth\DecoupledAuthUserInterface;
 use Drupal\flag\FlagInterface;
 use Drupal\flag\FlagServiceInterface;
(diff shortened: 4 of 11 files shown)


modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
```diff
@@ -0,0 +1,17 @@
...
+configuration:
+  email_subject:
+    value: 'Cannot subscribe to {{ entity_label }}'
+  email_body:
+    content:
+      value: |-
+        <p>Thank you for showing interest in keeping up with the updates for <a href="{{ entity_url }}">{{ entity_label }}</a>!</p>
+        <p>The email address you were using when trying to subscribe is already associated with a regular account on this website.</p>
+        <p>If you still want to subscribe to the updates made to the content, you can log in to the website, using your existing account, and then subscribe as a regular user.</p>
+        <p>If you do not want to subscribe or are unsure why you received this email, you can ignore this message.</p>
```
AaronGilMartinez, line 16: Message doesn't match exactly with requirements.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -154,20 +155,47 @@ public function buildForm(array $form, FormStateInterface $form_state, FlagInter
...
+    ];
+
+    // Check if a user with this email already exists.
+    // @todo Use dependency injection instead of function call.
+    $account = user_load_by_mail($mail);
+
+    if ($account !== FALSE) {
+      // The decoupled_auth module is installed, which replaces the class for
+      // user entities.
+      \assert($account instanceof DecoupledAuthUserInterface);
```
AaronGilMartinez: Remove assertion in production code, please

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
```diff
@@ -154,20 +155,47 @@ public function buildForm(array $form, FormStateInterface $form_state, FlagInter
...
+      if ($account->isCoupled()) {
```
AaronGilMartinez: No computing in between conditionals, so replace nested conditionals with only one, please.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -144,6 +155,33 @@ public function testSymfonyMailerOverrides(): void {
...
+      'registered_user_email_notice',
+      'Overridden subject for email taken',
+      <<<BODY
+<span>{{ entity_label }}</span>
+<span>{{ entity_url }}</span>
+BODY,
+    );
+
+    // Visit the article, and subscribe with a conflicting email address.
+    $this->requestSubscriptionForArticle('conflict@example.com');
```
AaronGilMartinez, line 162: This could be used in the same test when checking other mails.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -225,6 +281,50 @@ protected function doTestDefaultHtmlEmailContents(): void {
...
+   * Overrides an email template.
+   *
+   * @param string $template_id
+   *   Template id.
+   * @param string $subject
+   *   Overridden email subject.
+   * @param string $body
+   *   Overridden email body.
+   */
+  protected function overrideEmailTemplate(string $template_id, string $subject, string $body): void {
```
AaronGilMartinez: This function is only used once. We change it everywhere  in the test or we drop it.

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
```diff
@@ -209,6 +210,83 @@ public function testForm(): void {
...
+    ]);
+    $entity = $this->drupalCreateNode([
+      'type' => 'page',
+      'status' => 1,
+    ]);
+
+    // Create a regular user account.
+    $user = $this->createUser(values: ['mail' => 'conflict@example.com']);
+    $this->assertInstanceOf(DecoupledAuthUserInterface::class, $user);
+    $this->assertTrue($user->isCoupled());
```
AaronGilMartinez, line 230: Not needed `instaceOf()` assert, `isCoupled()` will be present on an instance of a class using that interface.

modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
AaronGilMartinez: This is adding `<p>` for text lines that are not paragrahs, for line breaks it's better to use `<br>`.

modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
```diff
@@ -96,9 +97,25 @@ public function confirmSubscriptionRequest(FlagInterface $flag, string $entity_i
...
+    }
+    catch (RegisteredUserEmailException) {
+      // The email address is already taken.
+      // The user has already proven at this point that they own the email
+      // address.
+      // It is therefore ok to reveal that a user account with this email
+      // address is associated with a regular account.
+      $this->messenger()->addWarning(
+        '<p>' . $this->t('You have attempted to subscribe as anonymous, using an email address that is already associated with a regular account.') . '</p>'
+        . '<p>' . $this->t('If you still want to subscribe to content updates for this item, you can log in to the website, using your existing account, and then subscribe as a regular user.') . '</p>',
```
AaronGilMartinez: Same here for `<p>`

modules/oe_subscriptions_anonymous/src/Plugin/EmailBuilder/AnonymousEmailBuilder.php
```diff
@@ -19,6 +19,7 @@
  *   label = @Translation("Anonymous subscriptions"),
  *   sub_types = {
  *     "subscription_create" = @Translation("Subscription confirmation creation"),
+ *     "registered_user_email_notice" = @Translation("Subscription confirmation - email address already taken"),
```
AaronGilMartinez:
```suggestion
 *     "registered_user_email_notice" = @Translation("Subscription confirmation - registered user email"),
```

modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
AaronGilMartinez: image

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -279,6 +379,31 @@ protected function assertConfirmMailHtml(EntityInterface $entity, string $mail_b
...
+   * Also asserts the CSS class in the outer wrapper, that derives from the
+   * mail key and module name.
+   *
+   * @param string $mail_id
+   *   Expected mail id/key within oe_subscriptions_anonymous.
+   *
+   * @return string
+   *   Trimmed mail body content without wrapper divs.
+   */
+  protected function getMailBodyWithoutWrapper(string $mail_id): string {
```
AaronGilMartinez, line 362: No need to filter by body > div, what we need to check is that is the correct mail type and get the inner content.
We can filter directly by the class. In fact for other mails we get inner content with: 
`$crawler->filter('.email-sub-type-user-subscriptions-access div.clearfix');`

Is the same, search the mail type class and get the div child with the clearfix class.

No need for a function for something that can be done in two lines:
`$crawler = new Crawler($mail->getHtmlBody());`
`$body = $crawler->filter('.email-sub-type-registered_user_email_notice div.clearfix')->html();`

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
```diff
@@ -209,6 +210,83 @@ public function testForm(): void {
     $this->assertTrue($pages_flag->isFlagged($page_two, $account));
   }
 
+  /**
+   * Tests a case where an email is already taken when a request is submitted.
```
AaronGilMartinez:
```suggestion
   * Tests a case where a registered user email is used on a anonymous subscribe request.
```

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
```diff
@@ -209,6 +210,83 @@ public function testForm(): void {
...
+   */
+  public function testEmailTakenOnSubscribeRequest(): void {
```
AaronGilMartinez:
```suggestion
  public function testRegisteredUserEmailSubscribe(): void {
```

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
```diff
@@ -209,6 +210,83 @@ public function testForm(): void {
...
+    $this->assertMailString('body', 'If you still want to subscribe to the updates made to the content, you can log in to the website, using your existing account, and then subscribe as a regular user.', $mail_data);
+    $this->assertMailString('body', 'If you do not want to subscribe or are unsure why you received this email, you can ignore this message.', $mail_data);
+
+    $mail_urls = $this->getMailFootNoteUrls($mail_data['body']);
+    $this->assertCount(1, $mail_urls);
+    $this->assertEquals($entity->toUrl()->setAbsolute()->toString(), $mail_urls[1]);
+  }
+
+  /**
+   * Tests a case where an email is taken when a reqest is confirmed.
```
AaronGilMartinez:
```suggestion
   * Tests a case where a registered user email is used on a anonymous subscribe confirmation.
```

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
```diff
@@ -209,6 +210,83 @@ public function testForm(): void {
...
+   */
+  public function testEmailTakenOnConfirm(): void {
```
AaronGilMartinez:
```suggestion
  public function testRegisteredUserEmailConfirm(): void {
```

AaronGilMartinez (changes requested):

modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
donquixote: The confluence page is here, https://webgate.ec.europa.eu/fpfis/wikis/pages/viewpage.action?spaceKey=openeukb&title=OEL+-+Subscription+-+Error+from+confirmation+link+in+the+email+for+corporate+email

The proposed text in concluence uses paragraphs.
Paragraphs provide more spacing, and make the email more readable.
Especially if the individual paragraphs already break to multiple lines due to viewport width or text width of the email viewer.

Besides the paragraphs, did I make a mistake somewhere?

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
donquixote: Using assert in this way is actually common in Drupal core.

Calling `assert()` is a middle ground between inline `/** @var ... */` docs and `if (...) {throw ...}`:
- Use `/** @var ... */` if you are sure this will always be the case, and you only need to tell static analysis about it (phpstan, or the IDE).
- Use `if (...) {throw ...}` or `if (...) {... special error handling }` if there is a legitimate possibility that this will fail, with conditions we cannot control within the component.
- Use `assert(...)` if the most likely cause of the fail is a programming error, that we would be able to see during development.

You can find more in this issue, https://citnet.tech.ec.europa.eu/CITnet/jira/browse/DQA-8651
Unfortunately, they did still not enable it in our docker images :(

There can be arguments for or against using `assert()`.
There might be an OEL convention that we never use them, in that case I can remove and either use `/** @var */`.

To me, `assert(... instanceof ...)` is almost always preferable over `/** @var ... */`, because it has the same verbosity, provides the same value for static analysis, but adds the additional (optional) runtime check in dev and testing env.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
donquixote: Yes, and in fact I had a branch which does that.
But then I was asked to produce the minimum pull request, which I did.

If we refactor other tests, I think it should be a separate PR.
I even proposed to create a "preparation" PR which does this kind of changes.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
donquixote: Same here. My other PR did use this everywhere.

Ofc what we could do is drop these kinds of functions for now, inline the code, and then do the full conversion in a separate PR.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
donquixote: The goal here was to have more information in failure messages:
If we receive an email with an unexpected class, we see a comparison showing the actual class found.

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
donquixote: We could use `/** @var .. */` doc instead of the assert.
The goal here is to let the static analysis (phpstan, IDE) know about the type, so it does not complain about the `->isCoupled()` call.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
donquixote: The reason for this two level if was to have the assert() in between.
We could drop the assert() and instead have a `/** @var DecoupledAuthUserInterface|false $account */` outside of the if(). Then it can be `if ($account !== FALSE && $account->isCoupled())`. It would be not so easy to do this with `assert()`.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
donquixote: E.g. the `$account instanceof DecoupledAuthUserInterface` would fail if we use a different version of the `decoupled_auth` module, which for some reason has removed this interface, or no longer replaces the user objects due to a misconfiguration somewhere. This can also happen in a website project that uses this module. In that case, the `assert()` produces an error in a dev environment where assertions are enabled. In a prod env without assertions enabled, we get instead a different error, where `->isCoupled()` method is not defined.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
donquixote: About "No computing in between conditionals", I am curious where this is coming from.
I can see how nested if can be something we want to avoid.
But I wonder where this specific "No computing in between conditionals" guideline or convention is coming from - personal experience, OEL convention, or something I would find published somewhere?
(This is more for the sake of learning than arguing.)

modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
donquixote: oops, is the `<p>` displayed with the message? In that case I should fix it.
I just checked in core, I don't see any place where a `<p>` is added in a status message. So seems this is not something meant to happen.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
AaronGilMartinez: As you posted needs configuration on dev images that is not ready yet, so has no use unless you do it manually.

We can use docummentary in the meantime for static analysis.

modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
AaronGilMartinez: Ticket description and linked confluence page in the ticket have different content than that one.

Regarding paragraphs, I don't see it in requirements, It's true its on tech analysis I didn't see it before. 

Still makes more sense to me semantically to use break lines but it's ok.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
AaronGilMartinez: Yes, this is for the case in which we drop the assert, really hard to do with it.

About where it came from is complexity, I have seen it in multiple places, for example mess detector, we are adding extra unnecessary levels.

Experience and OEL conventions are other, golden rule is avoid nesting if possible.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: Ok, then this can stay like this for the moment.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: This case is bit different, is a one use function, should be done in the future PR.

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
AaronGilMartinez: Doc is ok.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
AaronGilMartinez: In that case you can add an `assertCount(1, $crawler);`
With targeting class wrapper you expect instead of travelling, and asserting class.
Just an extra check.

modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
AaronGilMartinez: I think there are examples with other messages in the module.

modules/oe_subscriptions_anonymous/src/Form/AnonymousSubscribeForm.php
donquixote: Conclusion: Let's use `/** @var */` for now.

modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
donquixote: Actually, I realize that `$something . $this->t(..)` is a total anti-pattern in Drupal > 7 :)
`$this->t()` returns a markup object, and string concat converts it back to string, which we usually don't want.

donquixote: I see in other places we are using theme templates.
In this case I feel this would be overkill.
Any theme template we introduce becomes part of the API, and needs to be supported in future versions.

Alternatives:
- Put the `<p>` as part of the translatable string.
  In Drupal core I find 2 such places, both of them in `UserPasswordResetForm`.
- Use a wrapper `new FormattableMarkup('<p>@first_line</p><p>@second_line</p>', ...)`.
  For now this is the direction I would go with.

But now I notice a different problem.
If a page contains multiple message of same severity, when using Olivero or Claro, all of them appear in the same box, with no clear separation.

I get the idea that there is no clearly established best practice to put paragraphs or line breaks into status messages. Sad!

donquixote: In oe_whitelabel, such status messages are properly visually separated.
In the core themes they are not.
So we will go for `<br>`.

donquixote: I wanted to change the test to test the `<br>` in the message.
But I notice this would be too much trouble for now.
Let's keep the text-only assertion, and use a follow-up for test improvements.

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
donquixote: Line too long :)

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
donquixote: Line too long.
The general convention is that the first part of a doc comment should be a single line that is no longer than the line limit.

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
donquixote: Just to be clear, I much prefer the old name in most places, but I am going along with it :)

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
AaronGilMartinez: Ok

modules/oe_subscriptions_anonymous/tests/src/Functional/SubscribeTest.php
AaronGilMartinez: Ok

modules/oe_subscriptions_anonymous/src/Controller/SubscriptionAnonymousController.php
AaronGilMartinez: Thought about using existing template but indeed message only presents body, no title, adding the message with <br> seems the way to go.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
```diff
@@ -279,6 +347,34 @@ protected function assertConfirmMailHtml(EntityInterface $entity, string $mail_b
...
+    $content_div = $crawler->filter("body > div.email-type-oe-subscriptions-anonymous.$expected_class table div.clearfix");
+    if (!$content_div->count()) {
+      // The selector does not match.
+      // Show the actual email body, to make it easier to see what went wrong.
+      $this->fail(\sprintf(
+        "Expected an email body '%s' mail key. Found:\n%s",
+        $mail_id,
+        $this->email->getHtmlBody(),
+      ));
+    }
```
AaronGilMartinez:
```suggestion
    $this->assertCount(1, $content_div,
      \sprintf("Expected one email body '%s' mail key. Found:\n%s",
      $mail_id,
      $this->email->getHtmlBody()
    ));

**Review by AaronGilMartinez — CHANGES_REQUESTED** [review]

File: modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php (hunk 14 above)
Comment (donquixote) [reply modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php]: Yeah probably better.
Doing the `if (..) { $this->fail()` avoids the useless "failed asserting that 1 equals 0" line.
But better to stick with the common way of doing things.

**Review by AaronGilMartinez — DISMISSED** [review]

File: modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php [hunk 15]
```diff
@@ -279,6 +347,31 @@ protected function assertConfirmMailHtml(EntityInterface $entity, string $mail_b
...
+   *
+   * @param string $mail_id
+   *   Expected mail id/key within oe_subscriptions_anonymous.
+   *
+   * @return string
+   *   Trimmed mail body content without wrapper divs.
+   */
+  protected function getMailBodyWithoutWrapper(string $mail_id): string {
+    $crawler = new Crawler($this->email->getHtmlBody());
+    $expected_class = 'email-sub-type-' . \str_replace('_', '-', $mail_id);
```
brummbar: We don't use the \ for global scope functions.

modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
```diff
@@ -0,0 +1,17 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - oe_subscriptions_anonymous
+id: oe_subscriptions_anonymous.registered_user_email_notice
+configuration:
+  email_subject:
+    value: 'Cannot subscribe to {{ entity_label }}'
```
brummbar: This subject is strange. It follows the requirements, but I will have to challenge those. The message subject is still regarding the subscription to the entity, the user will be able to do it, just not with the anonymous system.

brummbar (changes requested):
Looks complete, just small comments.
Can you also drop from the commits the "Prep:" before the message?
Our commit standard is TICKET-NUMBER: MESSAGE.

modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
donquixote: If you want to talk with BA and propose a different subject, you have my support.
I was somewhat involved in the wording, but at some point just said whatever let's go.

modules/oe_subscriptions_anonymous/tests/src/Functional/HtmlMailsTest.php
donquixote: For reference here are some drupal.org issues.
https://www.drupal.org/project/drupal/issues/3346987
https://www.drupal.org/project/coding_standards/issues/3259666
These are still in progress and there is no final decision.

But here you are right if we have the internal convention we should stick to it, until we change it.

And tbh the main reason it's in this PR is due to how my IDE was configured - so laziness.
Going to fix.

donquixote:
> Looks complete, just small comments. Can you also drop from the commits the "Prep:" before the message? Our commit standard is TICKET-NUMBER: MESSAGE.

These prefixes preserve my sanity when I organize commits. It marks changes that do not provide the behavior change described in the ticket, but which are still kind of necessary as a preparation.
But ok to stick to OEL internal convention.
Will fix.

modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
donquixote: I will hear back from Lucas about this.

donquixote: So basically it's back to us.
For me this is ok.

Some ideas:
- "Please log in before you subscribe with this email address"
- "Anonymous subscription attempt with your account's email address"
- "Anonymous subscription attempt with registered email address"
- "Cannot use this email address for anonymous subscription"

There are two scenarios here:
- the legitimate owner of the email address who also has an account on the website tries to subscribe as anonymous
- somebody else tries to subscribe as anonymous and puts the "wrong" email here, impersonating the other user who already has an account.

donquixote: Ok, consensus is "Anonymous subscription attempt with registered email address".

AaronGilMartinez: Any subject using the term "anonymous subscription" should be avoided, this is a techinal term used for component naming or to be referenced in code by developers, not for final users.

Users aren't aware about roles in Drupal, or internal development naming used for each of the subscriptions.

AaronGilMartinez: As discussed and approved by consensus, the subject should be:
_Please log in to subscribe to {{ entity_name }}_

This gives enough information to the user that wants to subscribe, that a required action is needed in order to subscribe to the page.

Any specific project wording can be achieved by setup overriding the mail policies.

AaronGilMartinez (changes requested):

modules/oe_subscriptions_anonymous/config/mailer_override/symfony_mailer.mailer_policy.oe_subscriptions_anonymous.registered_user_email_notice.yml
donquixote: Actually it will be "log in", not "login".

AaronGilMartinez: Updated comment in favor of 'log in' instead of 'login' following Drupal core practice.

composer.json
```diff
@@ -80,11 +80,12 @@
...
             "composer/installers": true,
             "cweagans/composer-patches": true,
+            "dealerdirect/phpcodesniffer-composer-installer": true,
             "drupal/core-composer-scaffold": true,
+            "php-http/discovery": false,
             "phpro/grumphp": true,
-            "dealerdirect/phpcodesniffer-composer-installer": true,
             "phpstan/extension-installer": true,
-            "php-http/discovery": false
+            "tbachert/spi": true
```
AaronGilMartinez: Not needed

AaronGilMartinez (changes requested):

AaronGilMartinez (approved):

brummbar (approved):

PR #53, opened 2026-03-03
OEL-4512: Update to Drupal 11.3.x.

tibi2303:


tibi2303:
`composer self-update` was pulling Composer `2.9.x`, and that conflicts with `drupal/core-dev ~10.3.0`:
- `drupal/core-dev 10.3.x` needs `justinrainbow/json-schema ^5.x`
- `composer/composer 2.9.x` needs `justinrainbow/json-schema ^6.x`

This makes dependency resolution impossible for that matrix row.

Fix: pin CI to Composer `2.8.12` (`composer self-update --2.8.12`).

.github/workflows/ci.yml
```diff
@@ -19,11 +19,14 @@ jobs:
           - "10.4.0"
           - "11.1.0"
           - "11.2.0"
+          - "11.3.0"
         exclude:
           - php_version: "8.1"
             core_version: "11.1.0"
           - php_version: "8.1"
             core_version: "11.2.0"
+          - php_version: "8.1"
```
AaronGilMartinez: Is 8.1 supported still?

.github/workflows/ci.yml
```diff
@@ -40,15 +43,13 @@ jobs:
       - name: Start services with Docker Compose
         run: docker compose up -d
 
-      - run: docker compose exec -T web composer self-update
+      - run: docker compose exec -T web composer self-update 2.8.12
```
AaronGilMartinez: Why locking this specific version?

.github/workflows/ci.yml
```diff
@@ -19,11 +19,14 @@ jobs:
           - "10.4.0"
```
AaronGilMartinez: We may need to update 10 versions, these are not supported anymore. Lets confirm what versions we should support for all the components.

AaronGilMartinez (changes requested):

AaronGilMartinez:
Lets use centralised CI to sync versions tested.

If we drop a vesrion its good check for @todos

.github/workflows/ci.yml
donquixote: yep i see no good reason, did it break with latest?

.github/workflows/ci.yml
```diff
@@ -40,15 +43,13 @@ jobs:
...
-        run: >
-          docker compose exec -T web composer require
-          drupal/core:~${{ matrix.core_version }}
-          drupal/core-composer-scaffold:~${{ matrix.core_version }}
-          drupal/core-dev:~${{ matrix.core_version }}
-          --update-with-all-dependencies --ansi --no-progress --no-interaction
+        run: |
+          docker compose exec -T web composer require drupal/core:~${{ matrix.core_version }} --no-update --no-interaction
+          docker compose exec -T web composer require --dev drupal/core-composer-scaffold:~${{ matrix.core_version }} drupal/core-dev:~${{ matrix.core_version }} --no-update --no-interaction
+          docker compose exec -T web composer update --with-all-dependencies --ansi --no-progress --no-interaction
```
donquixote: If we already change this, the more elegant way is to use `composer update --with`, this way we don't overwrite anything in composer.json and we don't have to bother with --dev vs non-dev.

```
          docker compose exec -T web composer update --with drupal/core:~${{ matrix.core_version }} --with drupal/core-composer-scaffold:~${{ matrix.core_version }} ....
```

AaronGilMartinez (approved):

